PR #114, opened 2015-08-17
Household Rank rules text.

capitaladot:
I'v compiled the rules text for the Questoris Knight Household list.


Millicant:
Grabbed!  Thank you for doing this, it is very appreciated!

PR #122, opened 2015-09-02
Fix caestus

capitaladot:
Makes weapon profile from Predator with Magna-Melta shared between Caestus Assault Ram and Predator (was missing from Caestus).


capitaladot:
Hey @Millicant!

Millicant:
Whoops!  Sorry I passed over this by accident.  Thanks for the update!

PR #124, opened 2015-09-29
[Anon] File update: The Horus Heresy.gst

BSDataAnon:
**File:** The Horus Heresy.gst

**Description:** Removed erroneous modifier links to various Terminator units (and Pyroclasts) per:

Q: In the case of the ‘Pride of the Legion’ Rite of War, Veteran and Terminator squads are Troops choices in the force; could you clarify the exact squads which fall into these categories? A: In this case specifically this means Veteran Tactical squads and Legion Terminator squads only, so Justaerin and Deathshroud Terminators are not included in this. Any Legion-specific units in future Horus Heresy books which do qualify for this Rite of War will have this fact explicitly stated in their unit description. 
http://www.forgeworld.co.uk/resources/fw_site/fw_pdfs/Horus_Heresy/Betrayal_FAQ_Errata_V2.pdf


capitaladot:
... and it looks like I missed some modifiers to hide some of these squads when they were previously moved to troops. I'll commit this properly, now, on GitHub, but ignore this pull request, please.

capitaladot:
On second though, it looks like the Crusade Army List changed this entirely.

Millicant:
Correct, this issue is negated by the LA:CAL.

PR #133, opened 2015-11-02
Taghmata army list book.

capitaladot:
It is probably worth noting that I have shifted Ordos (where relevant) into the Game System File-provided "Legion" container.


capitaladot:
Got some testing done by the guys in the Mechanicum group on Facebook, and have incorporated their corrections.

capitaladot:
Implements #138

capitaladot:
I think this should be good to merge... @Millicant, any feedback?

Issue #140, opened 2015-11-06
Decapitation Strike


capitaladot:
Force Org for XIX Legion with 1 Heavy Support Slot.

PR #141, opened 2015-11-06
Decapitation strike

capitaladot:
Implements #140


capitaladot:
@Millicant  any feedback?

Millicant:
Sorry this one is going to be left out.  We'd have to add every single Force Organization chart modification for every single Rite of War and that is both inconvenient to code as well as use.

PR #151, opened 2015-11-23
[Anon] File update: The Horus Heresy.gst

BSDataAnon:
**File:** The Horus Heresy.gst

**Description:** When will you include codex chaos daemons in 30k? Obviously word bearers legion and cult militia can take daemons as battle/sworn brothers, also special daemo characters like Samus only works well in HH then standard 40k, because of its ruinstorm ability. I have a friend who talked with forge world about dark mechanicum, daemons and the affect of traitor legions and he said besides word bearers and cult militia, traitors are allies on convinance to daemons.


Millicant:
I think this was meant to be an issue, not a pull request.

Issue #164, opened 2015-12-28
Bugfix Magos Prime and Magos Dominus


BSDataAnon:
**File:** (HH) Mechanicum - Taghmata Army List (Complete).catz
- [x]  The Magos Prime has the redundant option to exchange their power weapon for a power weapon. Not redundant per se; included to present the the item as part of default equipment.

- [x] The "May take one second additional weapon" menu unlocked by selecting the Myrmidax High Order of Techno-Arcana for a Magos Prime will remain even if one subsequently changes the selected High Order of Techno-Arcana. 
- [x] No error is reported if these additional weapons are selected by a character who should not have access to them.

- [x] The Rad/Irad Cleanser and Photon Thruster are listed on the Magos Prime’s “May take one of the following” menu when they should be on the “May take one additional weapon” menu so that they can be taken a second time by a Magos Prime with of the Myrmidax High Order of Techno Arcana via the “May take one second additional weapon” menu. Currently it is impossible for a Magos Prime with of the Myrmidax High Order of Techno Arcana to select these weapons twice in Battlescribe even though they should have that option.

- [x] The Magos Prime has no option to take a Rad Furnace.

- [x] The Machinator Array is listed twice under the Magos Prime and Magos Dominus' "May take one of the following" menus.

- [x] None of the Magos Prime’s special rules are listed in its profile.

- [x] The Magos Dominus requires an entirely separate menu to allow access to one of a Rotator Cannon, Meltagun, Graviton Gun, Phased Plasma-Fusil, Rad/Irad Cleanser or Photon Thruster. Currently Battlescribe only allows the Magos Dominus access to the Rad/Irad Cleanser but it is located on the “May take one of the following” menu and thus disallows, or is disallowed by, the selection of other items of wargear that the rules should allow it to be taken alongside. The relevant options are present for the Legio Cybernetica exclusive (Legion selection) Archmagos Dominus.

- [x] The Magos Dominus is not listed as possessing the Patris Cybernetica or Feel No Pain (5+) special rules.

capitaladot:
https://github.com/BSData/horus-heresy/pull/167

PR #167, opened 2015-12-31
#164

capitaladot:
The Magos Prime has the redundant option to exchange their power weapon
for a power weapon. Not redundant per se; included to present the the
item as part of default equipment.

 The "May take one second additional weapon" menu unlocked by selecting
the Myrmidax High Order of Techno-Arcana for a Magos Prime will remain
even if one subsequently changes the selected High Order of
Techno-Arcana.

 No error is reported if these additional weapons are selected by a
character who should not have access to them.

 The Rad/Irad Cleanser and Photon Thruster are listed on the Magos
Prime’s “May take one of the following” menu when they should be on the
“May take one additional weapon” menu so that they can be taken a second
time by a Magos Prime with of the Myrmidax High Order of Techno Arcana
via the “May take one second additional weapon” menu. Currently it is
impossible for a Magos Prime with of the Myrmidax High Order of Techno
Arcana to select these weapons twice in Battlescribe even though they
should have that option.

 The Magos Prime has no option to take a Rad Furnace.

 The Machinator Array is listed twice under the Magos Prime and Magos
Dominus' "May take one of the following" menus.

 None of the Magos Prime’s special rules are listed in its profile.

 The Magos Dominus requires an entirely separate menu to allow access to
one of a Rotator Cannon, Meltagun, Graviton Gun, Phased Plasma-Fusil,
Rad/Irad Cleanser or Photon Thruster. Currently Battlescribe only allows
the Magos Dominus access to the Rad/Irad Cleanser but it is located on
the “May take one of the following” menu and thus disallows, or is
disallowed by, the selection of other items of wargear that the rules
should allow it to be taken alongside. The relevant options are present
for the Legio Cybernetica exclusive (Legion selection) Archmagos
Dominus.

 The Magos Dominus is not listed as possessing the Patris Cybernetica or
Feel No Pain (5+) special rules.


amis92:
After merging, please delete branches: helps keeping repo clean. Otherwise, currently there are 19 branches: https://github.com/BSData/horus-heresy/branches/active

PR #248, opened 2016-03-11
Re-implement the Caestus fix that was clobbered AGAIN.

capitaladot:


Millicant:
I'm not familiar - what was wrong?

PR #255, opened 2016-03-16
HH6: New Legions Added

Millicant:
With some possible errors/points corrections for wargear required


Millicant:
Closes #250 
Closes #251 
Closes #249

Issue #525, opened 2016-11-14
[Anon] Bug report: (HH) Legiones Astartes - Crusade Army List.catz


BSDataAnon:
**File:** (HH) Legiones Astartes - Crusade Army List.catz

**BattleScribe version:** 2.00.06

**Platform:** iPhone / iPod / iPad

**Dropbox:** No

**Description:** Unit: 
Proteus Land Raider

Issue:
No option to take hull mounted Twin Linked Lascannon

DonMorr:
Closed.

Issue #510, opened 2016-11-11
False Entries: Legion seeker


Nessaiy:
Kraken Bolt Shell -> Explosive is missing
Scorpius -> S5/Ap2 not 4/5 +Rending

DonMorr:
Fixed.

Issue #217, opened 2016-02-15
New Red Legiones Book (release date March 16th): rules and unit changes.


capitaladot:
LEGION UPDATES FROM THE NEW LEGION RED BOOK
IF A UNIT IS NOT MENTIONED, THEN IT IS UNCHANGED
 
**Emperor's Children**
- [x] All models with the LA rule gain +1 initiative on the charge
sonic shriekers are now -1WS to enemy models in base contact which aren't immune to fear. Effects the full length of the combat (not just first round), not just if you charge.
- [x] phoenix terminators are WS5
- [x] phoenix terminator sergeant now have 3 attacks
- [x] the cacophony weapons are heavy2, previously heavy1

**Iron Warriors**
- [x] iron circle added, elites choice, single unit can be taken as a command squad to perturabo. said unit gains a minor buff of their choice from a list.
- [x] tyrant seige terminator squads now heavy support choices
- [x] iron havocs are BS5

**Imperial Fists**
- [x] phalanx warder squads 'shield wall' rule fixed, now confers +1WS instead of +1I.
dorn's chainsaw has 'unwieldy' replaced with 'reaping blow' (see DG section)

**Night Lords**
- [x] 'from the shadows' confers 5+ cover save, use to be 6+
- [x] 'talent for murder' confers +1 to hit and wound, use to be just to wound
- [x] terror squads gain 'precision strike'
- [ ] sevatar has his additional psyker rules from the FAQ added to his profile, also gains precision strike

**Iron Hands**
no changes 

**World eaters**
- [x] 'blood lust' is now on a 4+ rather than a roll of a 1.
previous version of 'incarnate violence' replaced with may re-roll 1's to wound on the turn they charge

**Ultramarines**
no changes

**Death Guard**
- [x] may re-roll dangerous terrain tests, regardless of what the terrain is
- [x] instead of forcing successful rolls to wound with poison and fleshbane, DG get FNP(4+) instead
- [x] power scythes lose unwieldy and the ability to replace attack value with number of enemy models in base contact with the following
'reaping blow': -1 initiative, if in base contact with more than one model at their initiative step, they gain +1 attack
- [x] grave warden toxin rule replaced with 'poison(3+)' on death cloud and grenade launcher
- [x] deathshroud can be taken as a HQ or elites choice, plus as a command squad for a terminator praetor or mortarion. 
- [x] mortarion's scythe is no longer unwieldy, replaced with 'reaping blow'.
[...]

PR #538, opened 2016-11-20
Various fixes for AoD Army List:  #510, #217, #525, #540

DonMorr:
Various corrections to options and stats.


Millicant:
You'll have to explain this one a bit.  All those options for Terminator armour already exist.  What is changing?

Issue #517, opened 2016-11-13
Legiones Astartes - Rhino missing MultiMelta pindle mount option


zopha:
It has the rest, just doesn't have the MM option anymore (it did previously).

DonMorr:
Fixed.

PR #555, opened 2016-11-24
#517 Added Multi-melta to Legion Rhino and corrected other option costs

DonMorr:


zopha:
Tested on desktop client and all working.

Issue #568, opened 2016-11-28
Astarte Heavy Support changes


zopha:
A few changes for this slot. Dedicated transports next and then on to something more complicated :)

Heavy Support

**New Units -** 

- [x] LEGION JETBIKE SKY SLAYER SUPPORT SQUADRON - LA:AODAL P.60
- [x] LEGION MALCADOR ASSAULT TANK LA:AODAL P.81
----

**Points and Rule changes -** 

- [x] DEATHSTORM DROP POD - LA:AODAL P.61
  - [x] Deathstorm Frag Launcher - Change to Heavy 2
  - [x] Deathstorm Krak Launcher - Change to Heavy 3
----

- [x] LEGION HEAVY SUPPORT SQUAD - LA:AODAL P62
  - [x] Add Single Lighting Claw to "Exchange Chainsword/Combat Blade for"
  - [x] Flakk missiles available for every option, change so only available when Missile Launcher is selected (also change from Flak to Flakk)
----

- [x] LEGION PREDATOR STRIKE SQUADRON - LA:AODAL P63
  - [x] Command Tank option should only be available when 3 Tanks are selected
  - [x] Command tank points change (increase by 10)
  - [x] Heavy Conversion Beamer points change
  - [x] Pintle mount - Heavy bolter  and Combi-weapon points change - Add MM option 
  - [x] Extra Armour reduce by 5 points
  - [x] Hunter Killer reduce by 5 points
  - [x] Standard wargear
    - [x] Searchlight
    - [x] Smoke Launcher
----

- [x] LEGION LAND RAIDER BATTLE SQUADRON LA:AODAL P. 64
  - [x] Command tank points change (increase by 10) LA:AODAL P. 65
  - [x] Pintle mount - Changes points to correct value LA:AODAL P. 65
  - [x] All land raiders  Extra Armour reduce by 5 points- Hunter Killer reduce by 5 points LA:AODAL P. 65
  - [x] Achillies - Add Armoured Ceramite LA:AODAL P. 65
  - [x] Proteus - Add hull-mounted Twin-linked Lascannon LA:AODAL P. 65
  - [x] Achillies - Add Quad Launcher ammunition types and costs  (Incendiary shells, Splinter shells, Phosphex canister shot (Phosphex only with Siege Master consul) LA:AODAL P. 65
----

- [x] ACHILLES-ALPHA PATTERN LAND RAIDER LA:AODAL P. 67
[...]

PR #614, opened 2016-12-18
Fixes for Astarte Heavy Support - Basic units

zopha:
This is a complete fix for the basic Heavy Support units of Astartes's list, it covers all the entries in #568 and fixes them. I haven't made changes (much at least) to the other sections, rites or legion equipment. The aim of this fix is to get the basic units fixed and correct before moving on to another slot (like elites, troops, etc) eventually working up to HQ, rites and then legion modifiers. Some legion modifiers are in place from work carried out before me, so if your legion stuff is in and working (mostly Heavy Flamers strength change for salamanders and assualt cannon's for BA) then great! We'll get to the rest in future fixes.

Please reivew and let me know of any changes that need to be made! Either from a roster editor perspective or a data file perspective.

### DIFF
--- a/(HH) Legiones Astartes - Crusade Army List.cat
+++ b/(HH) Legiones Astartes - Crusade Army List.cat
@@ -61435,7 +61435,7 @@ Phalanx Breacher squads and Legion Terminator squads may be taken as troops in a
       </infoLinks>
       <modifiers/>
       <constraints>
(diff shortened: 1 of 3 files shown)


zopha:
Added fixes for
New Breacher points value
Fixed Libarian master levels (wasn't checking child selections in the Centurion entry)
Fixed bots for Praevian (wasn't checking child selections in the Centurion entry)
Added in hand flamer for Legion Vet squad with a condition that only displays it when BA legion is selected.

PR #644, opened 2017-01-09
Fixes astarte fast attack

zopha:
Remaining heavy support unit bugs, 
Added missing command squad for Shadrak
Added Leman Russell, Russ' tamer cousin who uses "Get you by" rules
Updated Fast Attack units (fixing points, missing profile, missing options and missing rules)

Updated units in the fast attack slot, corrections and optimisations for BS2.

BS2 optimisation still needst to be done for Outriders, Attack bikes, Jetbikes and Anvillus.

### DIFF
--- a/(HH) Legiones Astartes - Crusade Army List.cat
+++ b/(HH) Legiones Astartes - Crusade Army List.cat
@@ -27355,7 +27355,26 @@ Ignores Cover special rule.</description>
           </costs>
         </selectionEntry>
       </selectionEntries>
-      <selectionEntryGroups/>
+      <selectionEntryGroups>
+        <selectionEntryGroup id="c8be-2c4f-bdc8-36a7" name="Retuine Command Squad" hidden="false" collective="false">
+          <profiles/>
+          <rules/>
+          <infoLinks/>
+          <modifiers/>
+          <constraints/>
+          <selectionEntries/>
+          <selectionEntryGroups/>
+          <entryLinks>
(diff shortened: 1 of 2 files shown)


zopha:
The one posted on the web is the same as the one here and the one I have. We can give it a try or at least check it.

Are you using mobile or desktop? And, sorry have to ask, have you deleted your catalogue files and made battlescribe download them again?

When you're creating the roster you should have "The Horus Heresy v44" and Age of Darkenss Army List 205. If you don't have both these versions then the files aren't updating properly. Might be better to delete the files manually and then getting BS to download them again

If you've done that then the next thing to try checking the data file yourself. If you open the legion file with the data editor on desktop, expand the second last set of entries "selection entries" and scroll down until you see a chain link icon for the entries. Scroll down to the D's and you'll see three entries for the deathstorm. If you select the last one you should see what force slot the entry the entry belongs to, so you can see which is the fast attack entry. Expand that entry and you'll see the set hidden to true, expand that and you'll see the condition that keep it hidden unless there is a selection in the relevant entry menioned.

2017-01-09 19_20_18-data editor 2 00 09 - legiones astartes_ age of darkness army list

If you selected the "If equal to" you'll see the info on the right, it should be set to Orbital Assault and the Enter ID should be 8c4d-a5af-f886-05cc

2017-01-09 19_30_11-data editor 2 00 09 - legiones astartes_ age of darkness army list

PR #793, opened 2017-02-28
Alphalas

alphalas:


capitaladot:
Could we get some issue numbers or something in these?

alphalas:
If you look in the commit the not merge commit it should have a comment in there that says which issue I closed with it. I just type it in when I commit it to my branch and then I pull that to the master

Issue #825, opened 2017-03-08
Knight Bugs and War Machine Detachment Constraints


capitaladot:
LA should allow 1-2 of the same selection in a War Machine Detachment
The Porphyrion is a LoW for Questoris, not a hull choice for Household Rank units (those Knights in the other FoC slots).

PR #826, opened 2017-03-08
LA should allow 1-2 of the same selection in a War Machine Detachment

capitaladot:


capitaladot:
Check appropriate show/hide behavior for cybernetica units in Questoris.
Check that the Porphy is a LoW, not a Household Knight choice, in Questoris.
Check that 1-2 of the same LoW only is selectable in the LA list.

capitaladot:
Fixes #825

PR #875, opened 2017-04-14
Fixes the readme

capitaladot:
Various markdown was in another markdown flavor that no longer works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,13 @@ Horus Heresy
 ============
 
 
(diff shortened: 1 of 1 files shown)


zopha:
Did the markdown flavor change recently?

amis92:
@zopha it seems that there was introduced a stricter dialect. Almost all READMEs on BSData are broken, and many across whole site.

PR #1182, opened 2017-11-22
Fix SW compulsory troop hides, hunter killer mission, Damocles max

capitaladot:
- fix inverted logic that hid compulsory Troops
- fix Hunter Killer Mission and add rules text
- fix Damocles


zopha:
Space Wolves and troops selections look better now, thanks. Just Assualt squad needed a condition moved to the group conditions so the compulsory version will be hidden for a SW list. https://i.imgur.com/9lAJjXG.png

I see the change for the Damocles so it's hidden if you have less than 1000 points limit in the roster, it doesn't trigger if you build an unlimited list. I was trying to go with a constraint that would count the points currently in the list rather than the initial roster limit so it would show an error until you have at 1000 points added to the list.

Not familiar with Hunter Killer so can't comment much on it.

Just added a default weapon for the combi-bolter on a Varagyr Thegn.

Issue #1120, opened 2017-09-18
Issues with Solar Auxilia HQs


TomTheHand:
I've looked over the Solar Auxilia HQs and found some issues; I haven't looked at the rest of the Auxilia in this level of detail yet.

**Legate Commander (Crusade Imperialis, pg. 24)**
- [x] Hand Flamer, Inferno Pistol, Plasma Pistol, and Power Fist do not point to weapon profiles.
- [x] Currently it is not possible to select a specific type of Power Weapon.
- [x] The Volkite Serpenta does not cause the Deflagrate special rule to display (it does correctly list Deflagrate in its type).
- [x] Selecting an Iron Halo or Displacer Matrix does not remove the Refractor Field entry from the unit profile.
- [x] Selecting Digital Lasers does not add a descriptive entry to the unit profile.
- [x] Selecting Artificer Armour does not modify the Save in the stat line, add a descriptive entry to the unit profile, or remove the entry for Void Armour.
- [x] Perhaps upgrading a Legate Commander to Lord Marshal should change the unit name.

**Lord Mashal Ireton MaSade (Crusade Imperialis, pg. 25)**
- [x] MaSade's profile does not include his master-crafted Power Sword.
- [x] MaSade should be restricted to Loyalist allegiance.

**Auxilia Tactical Command Section (Crusade Imperialis, pg. 26)**
- [x] For the Strategos, the Hand Flamer, Plasma Pistol, and Power Fist do not point to profiles.
- [x] Currently it is not possible to select a specific type of Power Weapon for the Strategos.
- [x] Veteran Auxiliaries have individual entries for each model, and currently have their lasrifles implemented as a numeric entry that defaults to 1. Validation errors are thrown correctly if the user increases the number above 1, if the user selects a special weapon without setting it to 0, or if the user sets it to 0 without selecting a special weapon, but this is still a really wonky implementation.
- [x] For Veteran Auxiliaries, the Flamer, Meltagun, and Plasma gun do not point to profiles.
- [x] For Veteran Auxiliaries, the Volkite Charger does not cause the Deflagrate special rule to display (it does correctly list Deflagrate in its type).
  _Note that Veletaris Storm Sections and Household Retinue Squads, which also use Volkite Chargers, DO display the Deflagrate special rule)._
- [x] The Command Section should be able to be used as the Compulsory HQ.

**Auxilia Tank Commander (Crusade Imperialis, pg. 27)**

- [x] The user should be required to take an appropriate tank for each commander.
[...]

PR #1185, opened 2017-11-27
Ticked And all child selections on the constraint for Tank cetegory a…

zopha:
…s it wasn't counting nested selections.


zopha:
fixes #1120

PR #1250, opened 2018-02-09
Minor cal issues

zopha:
Fixing some minor issues and replacing profiles with ones from the GST.

### DIFF
--- a/The Horus Heresy.gst
+++ b/The Horus Heresy.gst
@@ -14339,7 +14339,7 @@ Command Benefits:
     </selectionEntryGroup>
   </sharedSelectionEntryGroups>
   <sharedRules>
-    <rule id="5342-bf92-d6c0-ebb9" name="Hit and Run" book="BRB 7th" hidden="false">
+    <rule id="5342-bf92-d6c0-ebb9" name="Hit and Run" book="HH: AoD Rulebook" page="164" hidden="false">
       <profiles/>
       <rules/>
       <infoLinks/>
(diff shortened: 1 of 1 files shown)


capitaladot:
Checking out now.

capitaladot:
Bug: Calibanite Warblade seems to cost the Apothecary 20 points.

zopha:
Not sure where the max1 came from for combi-weapons but I'll get look at these today. Thanks for finding them!

zopha:
Not seeing the issue on Seekers but something different, the Additional wargear is an Add Item box now.

Imgur

May as well replace it with model entries like the Veteran Tactical Squad if it's acting like this. I can't see the combi-weapon validation error, just the one it throws when first added making the player choose which combi-weapon they want.

I can't see the venom sphere cost issue on desktop but I can on mobile. Desktop is 2.01.09 but mobile is 2.01.16.... somethings changed. The data file has a modifier to set the entry to 0 points but it's not working for some reason on the units you mentioned. It doesn't matter that much tbh as we can move away from a Selection Entry with min1max1 for mandatory equipment and just put in a linked profile.

capitaladot:
I fixed the Venom Sphere thing.

zopha:
Fixed the additional wargear thing on seekers, it's now min1max1 and the sub-options are working.

Can you check the combi-weapon limit again? I've tried and can't replicate it on windows with these files and on android with the current release.

PR #1274, opened 2018-03-06
Closes #1269

### LINKED ISSUE #1269: Magos reductor paragon blade option does not display correctly. It appears to be set to hidden if no selection of archmagos is made, however reductor has a separate and bespoke selection of archmagos reductor instead of archmagos (shared) and it doesn't trigger as it should.
**File:** (HH) Mechanicum - Taghmata Army List.catz

**BattleScribe version:** 2.01.16

**Platform:** Android

**Dropbox:** No

**Description:** Magos reductor paragon blade option does not display correctly. It appears to be set to hidden if no selection of archmagos is made, however reductor has a separate and bespoke selection of archmagos reductor instead of archmagos (shared) and it doeesn't trigger as it should.

capitaladot:


zopha:
This is for the magos Reductor for #1269 ? The CAL file was changed though, looks like an edit to the Dragon Scale wargear option.

zopha:
I'm going to close this one off as the same fix is present in pull request #1275

Issue #1308, opened 2018-04-14
Mechanicum: Taghmata - Thallax Cohort missing stats for Multi-melta


eaxexe:
**File/Catalogue:** Mechanicum: Taghmata Army List v65

**BattleScribe version:** v2.01.10.102

**Platform:** Windows

**Dropbox:** Yes

**Description:** When adding a Multi-melta to a Thallax Cohort, in a Ordo Reductor, Matrix of Ruin, list, I am missing the stat-line for the Multi-melta

The Multi-melta is also missing under "Weapon"

PR #1309, opened 2018-04-14
Added missing stats to Ursarax and Thallax Cohorts

eaxexe:
Added stats for Power-fist to the Ursarax Cohorts and Multi-melta to the Thallax Cohorts

Solving #1306 and #1308


capitaladot:
Thanks @eaxexe!

PR #1325, opened 2018-04-15
Adding Terrax Pattern Termite Assault Drill for Mechanicum

eaxexe:


eaxexe:
Also fixed incorrect points for Myrmidon Destructors

PR #1465, opened 2019-01-02
Solar auxilia minor fixes

stvnksslr:
* Veletaris storm section had incorrect cost for per unit power axes they were 45pt per unit should only be 5
* auxilia rapiers didnt have any profile for the multi-laser on the unit
* multi-meltas for servo automata had no statline

This is my first PR for this so let me know if I have missed something thank you,


capitaladot:
Thanks, Steven, will review.

capitaladot:
So, Power Axes should be 5 points per model.

Rules quote:
The Section’s Veletarii may exchange their volkite chargers for one of the
following weapon types. If this option is chosen, all of the Veletarii must
be upgraded to carry the same weapon. If this is done then the Section’s
Prime may also exchange their weapon for the same type as the Section,
although this is not compulsory:
Rotor cannon
.................................................................................Free
Power axe
....................................................................+5
points each

Are you reporting they are being charged 45 points per model instead of the
correct 5 points per model?

stvnksslr:
@capitaladot 

In the current file you can change each veletaris to have a 45 point power axe, I changed them to work as it should be in the books and quoted rule above where you flip the whole squad for 5pts per unit so 45pts total for the squad.

The household retinue works as it should for reference.

capitaladot:
So, each refers to Model. Unit refers to the whole squad.

There is only supposed to be one "collapsed" set of Velentarii Models
listed (that's why it is plural), starting at 9x models, then incrementing
up.

This is what I see in my test roster on Android and Java (MacOS/Windows).

When I add a new Storm Section, I see multiple Velentarii listed, which is
the real bug.

It does look like the "collective" checkbox has been turned off on the
Velentarii entry, which broke this.

stvnksslr:
@capitaladot 

got ya that makes sense, ive changed it to line up more with that, though the default selection seems to not fire for some reason though ive seen this in other entries.

thank you,

stvnksslr:
@capitaladot Just bumping this in the mean time, please take a look when you get the chance I made changes more along the lines of what you requested so that collective behaves properly.

capitaladot:
@stvnksslr Sometime this week in advance of FAQ changes.

capitaladot:
@stvnksslr were these addressed elsewhere?

stvnksslr:
@capitaladot was going to fix the conflicts and then put it back up didnt want to have a pr sitting out that couldnt be merged.

stvnksslr:
@capitaladot 

Good to go take another look when you get the chance, thanks for the help!

Issue #1472, opened 2019-02-04
Some minor issues, and some others too.


Tomakokamikaze:
Armillus Dynat & Autilon Skorr: Should have Artificer Armor as default. (you have to select it for some reason)

Power Glaive: is showing Strength as "User/1" should be: "User/+1"
Also, power Glaive is available for bodyguards, and since they arent characters, it shouldnt be.
Rite of War: The Sagyar Mazan, not showing up for whites scars (just for Shattered Legions)

The Nemean Reaver: Has a Master Crafted Bolt Pistol that is NOT in his profile.

Qin Xa and Raldaron have had their rules released in the weekender, maybe we can put them up already? their pre-released profiles are annexed below.

![qin xa](https://user-images.githubusercontent.com/21247755/52190774-23723180-2828-11e9-804b-dc885596b8b0.jpg)

![raldoron](https://user-images.githubusercontent.com/21247755/52190786-2c630300-2828-11e9-9829-97d69b987a48.jpg)

Mayegelt:
Thanks for the report. Had a bit of a busy week with Worky and stuff after having gone to the Weekender. Was going to add the new characters in today anyway :) But thanks for catching the other issues.

I fixed Armillus Dynat & Autilon Skorr. Seems to have been a remnant of an older build that had them with an option to take it, rather than just included by default. So no selection needed now.

I have changed Power Glaive to User/+1

Redid the selections for power weapons to stop terminator command squads from taking Power Glaives. 

Rite of War: The Sagyar Mazan IS Shattered Legion only. as per the picture i linked.
![image](https://user-images.githubusercontent.com/20280410/52532424-dc4ed980-2d1c-11e9-90d2-356f52f0f141.png)

Added Nemean Reaver Master Crafted Bolt Pistol.

Qin Xa and Raldaron Added

Tomakokamikaze:
oh, than, also, Scimitar Jetbikes for bodyguards are not showing any weapons

Mayegelt:
@Tomakokamikaze as in not showing heavy bolter or something else?

capitaladot:
Can only take a Heavy Bolter.

Mayegelt:
@capitaladot i was wondering if he had found that a weapon option for the unit was not there. Latest update to my pull request has added stats for the Heavy Bolter in (of course salamanders can exchange for Heavy Flamers)

capitaladot:
Keen!

Mayegelt:
[...]

PR #1473, opened 2019-02-10
FAQ + Playtest rules, Various fixes and updates. Added Qin Xa and Raldoron

Mayegelt:
Issues from #1472 resolved

I fixed Armillus Dynat & Autilon Skorr not starting with Artificer Armour

I have changed Power Glaive to stats to S User/+1

Redid the selections for power weapons to stop terminator command squads from taking Power Glaives. 

Added Nemean Reaver Master Crafted Bolt Pistol.

Qin Xa and Raldaron Added


Mayegelt:
Added all FAQ and Playtest rules.
Playtest rules are accessable as a selection from Force Org (like HQ, Troops...)

capitaladot:
@Mayegelt can you rebase these as separate PRs? This is way too much to test at once.

thebiggeordiegeek:
Can’t find any issues with these myself, other than the lightning claw on the Sgt in Destroyers being free when it should be 10pts, but that’s an easy fix.

Mayegelt:
@capitaladot I could try to pull the thing apart, but it would be an absolute pain in the arse to do so as things like the Playtest rules stuff in some cases go with the non playtest stuff, and of course having put in a bit for playtest rules in of course affects all the playtest bits.

I have a few friends in the local group for me who have downloaded my version off my dropbox and no issues seem to be emerging outside of the couple of issues of Pts values (that have now been fixed as reported). But anything else can be sorted after i think.

Issue i guess is that as this was a big change i wanted to get as much done in 1 night as i could, as i have an upcoming tourney to go to, and was "in the zone" for doing something. Not sure how to split up into small sections, each as their own request. Especially as when i was making it, i started doing it 1 way, then switched as i found a better one part way through.

capitaladot:
@Mayegelt merged this, but let's try to keep separate things separate going forward.

capitaladot:
@Mayegelt I had to back this out, as it destroyed hours of work in the Centurion power weapons, and replaced tons of complicated show hide rules with incorrect show hide rules that were only applied to the Herald.

Issue #1502, opened 2019-04-08
Daemons of the Ruinstorm


martijndoornik:
With Book on its way, I will take on adding the Daemons of the Ruinstorm, unless someone else is already doing so. I will work on this this week and hope to have something ready to test this weekend. The list of things to do:

- [x] Adding the units;
- [x] Special Rules and such

- Applying the Emanations of Horror;
- [x] Bracket one;
- [x] Bracket two;
- [x] Bracket three;

- Applying the Aetheric Dominions
- [x] Crimson Fury;
- [x] Maddening Swarms;
- [x] Lurid Onslaught;
- [x] Creeping Scourge;
- [x] Mirror of Hatred;
- [x] Presplendent Terror;

Mayegelt:
I did a very basic start to it. Then got distracted. I had made all units and stats for them and points. But hadn't added emination, dominions, special rules.

martijndoornik:
Currently added the units (bar Unique units) and Emanations. Currently applying Dominions and such.

Track progress on https://github.com/BSData/horus-heresy/tree/daemons/initial this branch

Mayegelt:
Tested, but due to reimplementation of the old daemon codex and this one seemingly been created using it as a template, it either removes the new or old daemons from the current files and places the other in backup.

martijndoornik:
Fixed in the latest commit

PR #1504, opened 2019-04-09
Daemons/initial

martijndoornik:
## Description
Daemons of Book 8 added. 
**_The old Daemons codex catalog was removed_**

## Related Issues
#1502 

## Todos
Bigger rules (more text) have not been added, neither have the warlord traits, perils tables, etc. 

## Test Case
The entire daemon catalog as found on online boards can be used to create lists in games of the Horus Heresy.


Mayegelt:
Old daemons codex should remain active. In book 8 is says that that it is still able to be used.

martijndoornik:
Check, rejecting it then. Also hope to add Possessed later today. Will add those into a bigger pull request

Mayegelt:
Due to the way you modified the old daemons file and then have now readded the current one. It causes a conflict, that deletes the other file (or at least puts it into the backups folder).
This is due to them having shared elements in the unique ID numbers.

Issue #1511, opened 2019-04-17
[Anon] Bug report: (HH) Daemons of the Ruinstorm Army List.catz


BSDataAnon:
**File:** (HH) Daemons of the Ruinstorm Army List.catz

**BattleScribe version:** 2.01.21

**Platform:** iPhone / iPod / iPad

**Dropbox:** No

**Description:** Even after selecting the crimson fury aetheric Force selection, it still comes up as an error if I put Kabanda in the list, please fix this

Mayegelt:
Fixed in my lastest pull request. Will be posted soon

Issue #1510, opened 2019-04-17
[Anon] Bug report: The Horus Heresy.gstz


BSDataAnon:
**File:** The Horus Heresy.gstz

**BattleScribe version:** 2.01.21

**Platform:** iPhone / iPod / iPad

**Dropbox:** No

**Description:** When putting in two detachments both with the play test rules, it’s states that I can only have one plays test rule but then if I remove 1 it causes another error saying the detachment must have one, please fix this or remove it

Mayegelt:
Fixed in my lastest pull request. Will be posted soon

Issue #1509, opened 2019-04-16
Jump Packs giving wrong special rules


EricSmailys:
For many units, Jump Pack equipment appears to be adding the special rules Very Bulky and Relentless to the unit special rules. The correct rule for Jump Packs is Bulky, and they should not have Relentless (barring the unit getting the rule from something else).

Specific examples I've encountered, all while building a Blood Angels Loyalist list with the Day of Revelation Rite of War: Centurion (Chaplain) with Jump Pack selected (removing the jump pack also removes the special rules). Apothecaries with Jump Packs. Dawnbreaker squads. Of note, Angel's Tears squads (which come with Jump Packs by default) do not have this issue.

Mayegelt:
Fixed in latest update to be posted soon.

Issue #1508, opened 2019-04-15
(HH) Legiones Astartes - Crusade Army List issues


Toreador13:
Land Raider Phobos does not have a default hull mounted twin-linked heavy bolter listed as part of it's profile.  

Aether Rune armour has some odd issues in the Priest of Fenris selectors. When hovering over selection it shows the bolt pistol profile, and aether rune armour must be selected again after the first selection (probably more of an application issue. )  Should the profile of the priest of fenris show a wound increase with the aether armour selection? I see that it changes the armour save, but should W also be added?

Mayegelt:
Will be fixed in my latest pull request.

The stat increase from items like Aether Rune Armour and Digital lasers are not put onto the profiles of the characters by the application. We could do this for each case, but this could also cause confusion for those who are looking at the item description, so could end up with a stat profile upgrade then applying the upgrade again.

PR #1512, opened 2019-04-17
#1508 #1509 #1510 #1511 fixes

Mayegelt:
Fixed missing hull mounted heavy bolter on Land Raider Phobos #1508
Fixed missing rule entries for Jump Packs on some units. It seemed that the Jet Pack Entry for rules was being used for the Category Entry for them. #1509
Fixed Play test rule option when taking multiple forces (main and ally for instance) #1510
Fixed various Aetheric Dominion specific ‘Emanations of Horror’ Fixed Ka’bandha, as per issue #1511


Mayegelt:
Found issue to resolve

Issue #1520, opened 2019-04-22
Moritat Consul, they gain an additional Bolt Pistol that can be swapped out for other weapon options. In the Battlescribe file, they only have a single Bolt Pistol


BSDataAnon:
**File:** (HH) Legiones Astartes - Crusade Army List.catz

**BattleScribe version:** 2.01.24

**Platform:** Android

**Dropbox:** No

**Description:** When a Centurion is upgraded to a Moritat Consul, they gain an additional Bolt Pistol that can be swapped out for other weapon options. In the Battlescribe file, they only have a single Bolt Pistol and Close Combat Weapon when they should have 3 weapons according to their rules.

Mayegelt:
Same as #1518

In most recent pull request this is fixed. Just have to wait for Appspot to update if you don't manually do it from here. The reason a validation error was put in was we were getting 2-4 requests a week for "Why aren't X the new value" which are all part of the Playtest Rules, and as such are optional, rather than mandatory. So people needed to be given the option for the change, rather than enforcing it on them. And i had added playtest rules option to select, but people didn't seem to notice it.
I fixed it as i needed to switch it from roster to force (or the other way around i forget)

capitaladot:
This is a regression, as this has been dealt with at least twice previously.

Mayegelt:
Seemed to that it was all there, just a box for "Shared" on 3 different Modifiers were unticked.

PR #1521, opened 2019-04-23
#1520 #1523 #1524 #1525

Mayegelt:
Fixed issue with no secondary bolt pistol option. Required a tick in the shared box of 3 separate modifiers.


Mayegelt:
Fixes for
#1523 Contemptor-Incaendius limit
#1524 Bloodied Claws RoW
#1525 Macharius tanks Profiles

Also removed default selection of Power Weapon for Terminator Armour on Centurion, as this was defaulting on whenever any centurion was selected, meaning it was needed to be removed if you did not want Terminator Armour.

Issue #1533, opened 2019-05-03
[Anon] Bug report: (HH) Legiones Astartes - Crusade Army List.catz


BSDataAnon:
**File:** (HH) Legiones Astartes - Crusade Army List.catz

**BattleScribe version:** 2.02.01

**Platform:** Windows

**Dropbox:** No

**Description:** Assault Squad Sergeant, can take a pair of lightning claws, but this results in the roster being invalid, as you cannot take two + weapons. So although you can select pair of lightning claws, there is no option to replace the bolt pistol or combat blade/chainsword with nothing. There is no option to take nothing.

Mayegelt:
Modification of min/max settings to correct issue.

Issue #1530, opened 2019-04-28
Thousand son's missing data


zarmaglorg:
When creating a thousand son's heresy list on battlescribe I noticed that sekmet terminators don't seem to appear in either the elites slot or the troops when using Magnus

It would be appreciated if these entries could be added

Mayegelt:
They seemed to appear for me in Elites slot. But were not for the Troops slot with Magnus in the army.

Have corrected this, and also corrected issue of being able to take them as Elites in a PotL RoW. Have also corrected lack of Terminator Squads being Troops if Magnus is in the army.

Issue #1529, opened 2019-04-26
combi-weapon on sergeant of veterans cost twice.


Polnor:
I pay price x for combi-weapon, and also pay again when I choose what kind of combi-weapon I use.

capitaladot:
Somebody no doubt set a cost on the shared entry. We've gone round and
round on this so many times...

Mayegelt:
Fixed, though still looks messy with it showing as 0 pts to buy, but costing 10pts per one bought added to unit cost with no explanation.

PR #1534, opened 2019-05-05
#1529 #1530 #1533

Mayegelt:
#1529 Vet Squad Sergeant combi weapon fix.
#1530 Magnus / Sekmet / Termy fix for Thousand Sons
#1533 Assault Squad Sergeant Lightning Claws fix.


Mayegelt:
Just fixed #1535 as well. Seems to have been missing a selection option for the Playtest rules. Likely caused when the BS files were updated. Allies one was fine.
May need to look out for other army selection types that may have missed this option required.

PR #1546, opened 2019-05-26
Fixed involving Legion Astartes rule + Mine Plow fix

Mayegelt:
A couple of flags were set to "Or" rather than "And" as they needed to be.
This meant the LA must equal to or outnumber Tanks, Flyers and Vehicles was flagging as an error on all legions/row, rather than only on certain legions/row.
Also saw that Mine Plow was missing its description.


Mayegelt:
#1547 now also put in as a fix.

PR #1552, opened 2019-06-25
My local group has convinced me that Dedicated Transports count as Un…

capitaladot:
…its for Army Building, in regards to rules like Flesh over Steel, Host of Angels, and Ironwing Protocol.


Mayegelt:
Host of Angels has "Host of Angels: With the exception of Dedicated Transports,,,"
Thus already precludes that Dedicated Transports do not count towards the number of units that count towards the number of Vehicles counted in the army.

I have tested this pull request and it seems to entirely remove the limitation placed. I could have an army of contemptors or land speeders in a blood angel army and is allows it as valid.

Haven't tested other variants, but i know this one is broken.

Issue #1549, opened 2019-06-18
couple of issues with some Thousand Sons Characters


BSDataAnon:
**File:** The Horus Heresy.gstz

**BattleScribe version:** 2.02.04

**Platform:** Windows

**Dropbox:** No

**Description:** I have a couple of issues with some Thousand Sons Characters

1) I still get the "minimum 1 compulsory HQ" when Magistus Amon is selected, he is not a support HQ so this shouldn't happen
2) When I give Azek Ahriman a command squad, I should be able to give them mastery level 2 as an option but it is missing.
3) I should be able to upgrade the chaplain's power weapon (the one he comes with) with a force weapon but I can't
4) When I give my terminator librarian a pair of lightning claws, I still get the following errors: "Librarian must have one selection from force weapon", "Librarian must have one selection from exchange power weapon for", "Librarian must have one selection from exchange combi-bolter for". For some reason when I select the pair of lightning claws it doesn't get rid of these errors

Mayegelt:
Will fix issues 1 & 2.

But as for 3 & 4 the FAQ made it that the librarian was no longer able to swap the Crozius for anything else.

QUOTE
Legion Chaplain Consul (Page 20) Replace the text under the bullet point  Wargear to read: Replaces their chainsword or combat blade (or power weapon if in Terminator armour) with a Crozius Arcanum, which is a power weapon (type as shown on model). The Crozius Arcanum is a Chaplain’s symbol of office and may not be replaced by any other weapon.

tharcourt:
Your point on #3 is well taken, I had not noticed that in the FAQ

However that is not relevant to #4, as this is a librarian, not a chaplain, and has not crozius

Thanks for the help, it is much appreciated!

tharcourt:
I just wanted to follow up with this as these issues have still not been fixed. Thanks, appreciate the help!

Mayegelt:
Sorry for delay. I'll try to get the fix up soon. Had a bit of a bad time mentally recently. So not really been doing stuff.

Mayegelt:
Ok. Just about to upload changes.
Should be fixed.
Seems like only the termy squad was missing brotherhood of sorcerers thing, and no idea why Amon was missing the compulsory HQ thing.
Also fixed another error that was flagging up that Magnus, Ahriman and Amon were all not counting as the armies min requirement of Psyker lv 1.

PR #1553, opened 2019-07-13
#1549 Thousand Sons, #1556 DotRS, #1557 Talons fixes

Mayegelt:
1) I still get the "minimum 1 compulsory HQ" when Magistus Amon is selected, he is not a support HQ so this shouldn't happen.
FIX - Amon was missing the compulsory HQ tick.

2) When I give Azek Ahriman a command squad, I should be able to give them mastery level 2 as an option but it is missing.
FIX - Termy command squad was missing brotherhood of sorcerers,

FIX - Also fixed another error that was flagging up that Magnus, Ahriman and Amon were all not counting as the armies min requirement of Psyker lv 1.


Mayegelt:
Added Talons Shield Captain Tribune army building fix.

Mayegelt:
Just sorted the Greater Daemon Beasts issue.
Also had a play with the way it works things out to give a better way to display. Also corrected a Samus issue.

Issue #1572, opened 2019-09-23
New Moirax weapons


Mayegelt:
Add the 2 new Moirax weapons. Grav & Conversion.

Mayegelt:
Added last night. Had a few friends test. Seems to work

zopha:
Closed by #1574

Issue #1571, opened 2019-09-20
[Anon] Bug report: The Horus Heresy.gstz


BSDataAnon:
**File:** The Horus Heresy.gstz

**BattleScribe version:** 2.03.02

**Platform:** iPhone / iPod / iPad

**Dropbox:** No

**Description:** Can’t get any of the RoWs to populate. It shows it as red but nothing is there.

Mayegelt:
You must select a character that has "master of the legion" or "rite of command" before it will allow you to take a RoW.

zopha:
Closing, no response.

Also a change has been made on how this option is presented.

Issue #1570, opened 2019-09-10
 Maru Skara - Champion for compulsory second HQ


BSDataAnon:
**File:** The Horus Heresy.gstz

**BattleScribe version:** 2.03.02

**Platform:** Android

**Dropbox:** No

**Description:** When selecting the rite of war maru skara for emperors children i am not getting an option to select a champion for a co1mpulsory second HQ.

Mayegelt:
Had a quick look at this, and what was happening, is that it wasn't making you take a 2nd compulsory HQ. It was just insisting you take a Champion.
So previously you could have just taken the Champ and Fulgrim and it would have allowed it. This should not have been the case. So now you MUST take the Champ as 1 compulsory HQ and also take the additional Compulsory HQ of your choice.
Will post the fix when i'm done fixing other bits to go with it.

zopha:
Closed by #1574

PR #1574, opened 2019-09-25
Various Fixes, including RoW, Moirax and Thrall units

Mayegelt:
#1570 Mara Skara fixed, issued was allowing use of illegal lists
#1571 RoW no longer hidden before a Master of Legion is chosen. But will now just flag an error that one will be required.
#1572 New Moirax weapons added.
Additionally an issue for Thrall units for Mech was discovered and fixed relating to revenant alchemy.

Fixed issues from previous pull request, was throwing up an unexpected issue, now solved and remade.


Mayegelt:
@zopha this one should work

Issue #1577, opened 2019-09-30
[Anon] Bug report: (HH) Legiones Astartes - Crusade Army List.catz


BSDataAnon:
**File:** (HH) Legiones Astartes - Crusade Army List.catz

**BattleScribe version:** 2.03.03

**Platform:** Android

**Dropbox:** No

**Description:** There is an error saying too many selections for master of legion minimum 0.  You must have 1 to use rite of war

Mayegelt:
Seems to be an issue if you select the RoW first then a Master of the legion. On reloading the list it will unflag the error, so long as you have a MotL.

Mayegelt:
Looks like i have found a possible fix. Will try to use on my phone and post later today if working.

bluesif:
I am still getting this error even after trying your work around

capitaladot:
LOL, it isn't a workaround, you can't take a Rite of War without master of
the legion

bluesif:
Yes i know that....a work around in terms of the software giving an error for having MoTL

bluesif:
Issue was that my list wasnt at 1000 points yet, making MoTL illegal.  Please close case.

Mayegelt:
FIXED
Needs closing

Mayegelt:
Fixed ages ago. So can be Closed @capitaladot @zopha

PR #1578, opened 2019-10-01
RoW fix, fixed & Kyr Vhalen fixes

Mayegelt:
Issue #1577 occurred. This was more an annoyance than anything else. Reloading the roster after finishing fixed the issue. But made a tweek that seems to have fixed it. Not entirely sure if any other issues may arise from this, but shouldn't do.


Mayegelt:
Kyr Vhalen fixes added to request. Various issues such as he didn't have Warsmith by default and same for Servo Arm and stuff.

Mayegelt:
Added in a load of Militia fixes, mainly issues with lack of profiles for weapons.

PR #1596, opened 2019-10-30
Eldar & Necrons (Fanmade), + a couple of fixes.

Mayegelt:
This adds the Eldar and Necrons Fanmade rules from the AUS30K guys.
I've tested it a fair bit and seems to be working great

Any issues with this file i will resolve. Just leave a message for me to look at and i should be able to fix them.


Mayegelt:
Ok, a couple of fixes as referenced also included in this patch.

ALSO added Necrons, found the thing that was causing it to bug out. But not sure why it was bugging out with it. Readded the bit as a new one and it stopped complaining and crashing.

Issue #1602, opened 2019-11-07
[Anon] Bug report: (HH) Legiones Astartes - Crusade Army List.catz


BSDataAnon:
**File:** (HH) Legiones Astartes - Crusade Army List.catz

**BattleScribe version:** 2.03.10

**Platform:** Android

**Dropbox:** No

**Description:** With Legions Astartes: Dark Angels, when taking a Calibanite War Blade on an Assault Squad Sergeant, it is listed as S1 AP3, instead of S+1 AP3.

Mayegelt:
This is now included as a fix in todays update. Hopefully we be live soon.

Issue #1601, opened 2019-11-07
[Anon] Bug report: (HH) Legiones Astartes - Crusade Army List.catz


BSDataAnon:
**File:** (HH) Legiones Astartes - Crusade Army List.catz

**BattleScribe version:** 2.03.10

**Platform:** Android

**Dropbox:** No

**Description:** Iron warriors: Erasmus Golg is missing the combi-melta data in his entry

Mayegelt:
This is now included as a fix in todays update. Hopefully we be live soon.

PR #1603, opened 2019-11-13
Fixes

Mayegelt:
Necron fix for Eternity Gate Portal. For some reason was left as uncategorised (hopefully this will be the only fix needed) 
#1602 Calibanite War Blade fixed to +1 Strength, rather than just 1 Strength.
#1601 Iron warriors: Erasmus Golg added Combi-melta data in his entry


Mayegelt:
Hold off on pulling this one and I'll add the Deredeo stuff soon.
If not expect another pull soon after

Mayegelt:
Just put in Deredeo weapons for #1604 

Also removed a couple of duplicate entries in GST file, like Heavy Stubber, and corrected other files to cope with it.
Also noticed many many many entries for publications, all listed as their own ones. So removed various ones and corrected each army list to match. Fixed a few. Will work on others as and when I get time.

Can be pulled now. Will do more as i go.

Issue #1610, opened 2019-11-23
Legiones Astartes: Crusade Army list


Toreador13:
Thousand sons Praetors with psychic powers no longer can replace or add force weapon for 5 points when a power weapon is chosen, unless I can't find it somehow. It seems to be with the second (and) (or). If the second or with the specific leaders is removed the force weapon shows back up. (Praetor, Centurion, wolf lord, consol). I didn't get too much further into it as I am not sure how this data file is setup. But the root of the issue seems to be in there. 
![image](https://user-images.githubusercontent.com/6875311/69469202-2efdfd00-0d55-11ea-8900-3470050f8fd6.png)

Mayegelt:
Fixed

Issue #1609, opened 2019-11-22
[Anon] Bug report: The Horus Heresy.gstz


BSDataAnon:
**File:** The Horus Heresy.gstz

**BattleScribe version:** 2.03.13

**Platform:** Android

**Dropbox:** No

**Description:** Assault Squad additional wargear is not charging for power weapons. It should be 10pts each instead of free.

capitaladot:
Mistaken report

Toreador13:
Confirmed it is charging, but mine is adding a power axe ot the squad as additional wargear by default. 
![image](https://user-images.githubusercontent.com/6875311/69443587-51bbf180-0d14-11ea-947b-cbe7618988dc.png)

Toreador13:
Selecting 2 doesn't change the points cost either. 

![image](https://user-images.githubusercontent.com/6875311/69443787-b8410f80-0d14-11ea-8f00-161baee2de80.png)

Mayegelt:
Fixed

Issue #1608, opened 2019-11-21
[Anon] Bug report: (HH) Imperialis Militia and Cults Army List.catz


BSDataAnon:
**File:** (HH) Imperialis Militia and Cults Army List.catz

**BattleScribe version:** 2.03.08

**Platform:** Windows

**Dropbox:** No

**Description:** The option for Power Weapon(Sword, Axe Maul, Lance) for Ogryn Brute Squad should be 20 points not 10 points.

Mayegelt:
Will get on to it soon. Likely over weekend with any luck.

Issue #1607, opened 2019-11-19
Legiones Astartes General Bugs


Tomakokamikaze:
*Bugs:

Two handed/Heavy Chainsword -> no Weapon Profile shown

Assault Squads -> coming with power axes as additional gear as default

Legion Rapier Battery -> not showing Quad Heavy Bolter weapon profile

Pride of The Legion Rite of War -> Ebon Keshig not counting as compulsory troop choices

Blackshield Marauder Squads -> always showing in troop choices

Alpha Legion Headhunter Kill Teams -> should lose their Infiltrate when they pick a rhino as a dedicated transport (unless this rule was changed somewhere)

Alpha Legion Rewards of Treason -> General update and remove rules that are Legiones Astartes (X) and use Legiones Astartes (Alpha Legion)

*Suggested Improvements:

Almost all Units -> Should already come with the Chainsword/Combat Blades that are present in its standard profile

White Scars Kyzagan Assault Land Speeder -> I believe we can drop the "White Scars" and Just call it "Kyzagan Assault Land Speeder"

The Ebon Keshig -> missing Max Grenade Harness & Max Power Glaive
-> shouldn't come with a grenade harness as default
->Should come with 5 power glaives as default

Headhunter Killteams ->  Headhunters and Sargent should come with the combi weapon > combi bolter with banestrike rounds chosen as default

Mayegelt:
I'll have a check on these hopefully in the next week or 2. Got issue with my eye currently, so hard to do a lot though.

However
"Pride of The Legion Rite of War -> Ebon Keshig not counting as compulsory troop choices"
The unit has the Support Unit rule. So can't be compulsory, regardless of RoW or FoC.

"White Scars Kyzagan Assault Land Speeder -> I believe we can drop the "White Scars" and Just call it "Kyzagan Assault Land Speeder""
It could be dropped, but it's fine as it is imoo as it's the unit entry name in the book, like "The Kakophoni of the Emperor's Children" is listed like that, because it's the unit name in the book.

Tomakokamikaze:
Also, would be nice to have warlord trait selection options, but that is a bigger job

Mayegelt:
It's not a major issue to have warlord traits as a thing. I couldn't decide how to put it in though.
[...]

PR #1611, opened 2019-11-23
More Fixes for issues 16XX 06-10, 12-13, 15-16

Mayegelt:
#1606 
#1607 
#1608 
#1609 
#1610
#1612 
#1613 

Only issue i still seem to have is that Blackshields stuff is showing up, even in regular lists. Its rather odd because there is a hidden function that should be applying to them, but for whatever reason isn't. In addition this is also causing the legion list to not hide when blackshields are taken.


Mayegelt:
Add fix for #1612 and #1613

Mayegelt:
Added #1615 #1616

Mayegelt:
Should be good to go @capitaladot

PR #1638, opened 2019-12-31
1632, 1633, 1634, 1635 1639

Mayegelt:
BSData#1632 Knight Acastus knight Asterius added to gst, and also put into the tagmata, questoris and crusade in the LOW formations or Knight armours sections. Also moved the Karacnos Mortar Battery into GST and relinked it to the Tagmata Karacnos tank.
BSData#1633 fixed one of my mess ups with redoing the weapon choices for Mor Deythan.
BSData#1634 fixed pair of Ravens Claws selection for Centurions for both Arty and Termy armour, was previously not working and throwing up errors of to many selections or must still select power or combi weapons.
BSData#1635  Damocles command rhino now has options for Blessed Aurosimulacra for Iron Hands, and Shrapnel Bolts for Iron Warriors if it has a heavy bolter.


Mayegelt:
Added #1639 Daemons, Asuryani & Necron LOW Fix.
Needed a fix for the Lords of War from Imperial factions showing in each of those 3 lists.
Fix made by adding a new "Aliens and Daemons" selection to GST, then having a "Hide" function to all the LOW that if it is is in force then it hides them all.
Then added a min/max 1 selection of it into the Crusade (or Nodel Formation in the case of Necron).

PR #1640, opened 2019-12-31
Mechanicum Profile Fixes

stvnksslr:
Description
* First issue: Magos prime, if you give him a melta gun, the weapon profile doesn't show up.
* the same goes for the magos dominus, so i think something's wonky with melta gun in particular
* the archmagos dominus still can't choose a specific power weapon. just "power weapon"
* the hardened armour on the abeyant is missing its description
    apparently the rules for hardened armour can be found under the "breachers" entry in the legions book
* Ursarax are missing their lightning claws
* Castellax don't have the atomantic shielding listed properly

### DIFF
--- a/(HH) Mechanicum - Taghmata Army List.cat
+++ b/(HH) Mechanicum - Taghmata Army List.cat
@@ -4238,6 +4238,7 @@ Reduces transport capacity to 8.</description>
         </entryLink>
         <entryLink id="8f96-0dba-f6cb-ec83" name="New EntryLink" hidden="false" collective="false" import="true" targetId="0a18-686c-1dfb-4b71" type="selectionEntry"/>
         <entryLink id="f45c-8e8d-bfe0-0dd7" name="New EntryLink" hidden="false" collective="false" import="true" targetId="c36a-4017-0eb5-c8d2" type="selectionEntry"/>
+        <entryLink id="9cd0-475b-7369-3013" name="Karacnos mortar battery" hidden="false" collective="false" import="true" targetId="81f0-cd7e-d2fe-e7a0" type="selectionEntry"/>
       </entryLinks>
       <costs>
         <cost name="pts" typeId="points" value="225.0"/>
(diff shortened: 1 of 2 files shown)


(HH) Mechanicum - Taghmata Army List.cat
```diff
@@ -4234,7 +4238,6 @@ Reduces transport capacity to 8.</description>
         </entryLink>
         <entryLink id="8f96-0dba-f6cb-ec83" name="New EntryLink" hidden="false" collective="false" import="true" targetId="0a18-686c-1dfb-4b71" type="selectionEntry"/>
         <entryLink id="f45c-8e8d-bfe0-0dd7" name="New EntryLink" hidden="false" collective="false" import="true" targetId="c36a-4017-0eb5-c8d2" type="selectionEntry"/>
-        <entryLink id="9cd0-475b-7369-3013" name="Karacnos mortar battery" hidden="false" collective="false" import="true" targetId="81f0-cd7e-d2fe-e7a0" type="selectionEntry"/>
```
Mayegelt: Need to change this one back from my latest request

(HH) Mechanicum - Taghmata Army List.cat
```diff
@@ -10573,7 +10590,6 @@ Alternatively, instead of firing any weapons in the Shooting phase, the Magos/Ar
             <entryLink id="9291-22af-b60c-8f0b" name="Questoris Knight Warden" hidden="false" collective="false" import="true" targetId="f4f1-dad6-0596-ca5a" type="selectionEntry"/>
             <entryLink id="f73d-6c6a-8a8a-9dd9" name="Cerastus Knight-Lancer" hidden="false" collective="false" import="true" targetId="ca45ca55-eae1-1a81-8afc-1330611055a6" type="selectionEntry"/>
             <entryLink id="593f-5de1-07e1-a4cb" name="Acastus Knight Porphyrion" hidden="false" collective="false" import="true" targetId="4ded-9de3-f964-33a7" type="selectionEntry"/>
-            <entryLink id="d060-75e7-6ceb-e4ce" name="Acastus Knight Asterius" hidden="false" collective="false" import="true" targetId="72ad-abe8-94d8-5926" type="selectionEntry"/>
```
Mayegelt: Same with this one, needs to be kept as my current

stvnksslr:
Fixed no idea how those entries got effected BS data editor is really weird sometimes.

Mayegelt:
Had a look at it, and found a few extra missing bits, like the photon thrusters not having "Gets Hot" and plasma pistol option having no profiles at all.
I've merged the entry into my own one, and will be doing some more enhancements hopefully in the next couple of days.
Also played with Hardened Armour (moving it to GST and then relinking in Mech and LA, but think i need to change it from a rule to a wargear, as it will then show on units a bit more obviously.

Mayegelt:
I've also had a play with the lightning claws bit you added. So that it is the same as any multi weapon option one. So will then only show the item if you have it, but will also say the max restriction based on unit size (basically copy the formular from the power fist one, but also having to repeat it higher up. As at some point they may end up getting extra weapon options. Also once the Min/Max things that aren't working in BS currently. It will also make sure the person selects all the weapons for the unit.

image

stvnksslr:
@Mayegelt awesome thank you good catch on the lightning claws stuff, Ursarax also have a 5+++ fnp not a 6+++ like the lorica thallax gives, going to try and fix that in here shortly.

Mayegelt:
Cool. I have merged yours into my one i am about to post. So this one can be closed with that as well without needed to be pulled, as it will just replicate the work.
Most of it i just imported what you had done, apart from the alterations posted above. Very good work though. Saved me a load of time with various ones, as i could just copy paste them across onto my latest pull, as i was using them for my base file. As i thought they would have been pulled earlier.

Issue #1659, opened 2020-02-16
Minotaur points issue LA


Mayegelt:
LA version of the Minotaur was costing 275pts more than it should (should only be 205, was costing 480)

Mayegelt:
Pulled

PR #1660, opened 2020-02-16
#1659 #1661, #1662 and Inner Circle & Interemptors

Mayegelt:
LA version of the Minotaur was costing 275pts more than it should (should only be 205, was costing 480)


capitaladot:
This was probably the old pre Red Rulebook cost (where it was taken, as directed, from IA1e2) before the consensus emerged to take units from other red army list books since the IA series went out of print.

Mayegelt:
Dark Angels Interemptors added.

Mayegelt:
@zophia @capitaladot any chance of a pull?

Issue #1665, opened 2020-03-09
Primaris Nullificator forced to pay for Cataphractii Terminator Armour.


TheRavingRaven:
A centurion with the Primaris Nulificator upgrade is meant to get the Cataphractii terminator armour for free but is having to play points for it in the app. 
It needs to be edited so that Cataphractii armour is reduced to being free when the Primaris Nulificator Consul upgrade is selected.

Mayegelt:
Thanks for this one, will get on that one now.

Mayegelt:
In latest release

Issue #1661, opened 2020-02-19
[Anon] Bug report: (HH) Legiones Astartes - Crusade Army List.catz


BSDataAnon:
**File:** (HH) Legiones Astartes - Crusade Army List.catz

**BattleScribe version:** 2.03.12

**Platform:** Windows

**Dropbox:** No

**Description:** In the Legiones Astartes army lists, all the LoW superheavy tank variants of Baneblade, Shadowsword etc. lack of weapon and armor options. Also the list won't show all the weapons that the tanks have, e.g Baneblades hull mounted Demolisher cannon and Shadowswords hull mounted twin linked heavy bolter. I looked all those tank variants and their weapon/armor options from the HH: Conquest and Solar Auxilia army list(Battlescribe) and those were fine. If I'm wrong or I am missing something in the Legiones Astartes army list rules, ignore this message. I'm also sorry for my bad English and this not-so-clear message!

Mayegelt:
You are probably right. For some reason the ones in all the other armies are done differently to how they were added to legions.
I'll try to have a look. Though I think the weapons are there. They just aren't listed as separate selections in the rules for them, but are just rolled in as links

capitaladot:
The reason is the age of the red books. Only the Legiones Astartes ones
have ever been revised, and it was the second edition of those
(specifically Age of Darkness Army List) that first clarified taking random
other Forge World lords of war. Before the Age of Darkness Rulebook, it was
unclear if rules in Age of Darkness Army List about Lords of War applied to
other army lists from other books. The whole thing is a garden variety
GW/FW sloppy mess.

PR #1668, opened 2020-03-09
#1665 & #1670 & #1671

Mayegelt:
Had to reduce the cost of Primus Nullificator  Consul upgrade by 35pts as the Consul has Cata armour by default and the way the armour gets its value seemed unable to then be modified by the consul being taken. So the reduction has had to be factored in on the consul itself to balance it out.


Mayegelt:
Another new weapon for the Arquitor

Also noticed that for some reason i'd put side armour as only 11, it should have been 12. Might be the older sheet had lower armour, or i may have made a mistake.

capitaladot:
> Had to reduce the cost of the Consul upgrade by 35pts as the Consul has Cata armour by default and the way the armour gets its value seemed unable to then be modified by the consul being taken. So the reduction has had to be factored in on the consul itself to balance it out.

It had at one time been set to 0 then modified, but someone re-did it.

What consul gets free Cataphractii?

Mayegelt:
Just the Primus Nullificator. @capitaladot

Mayegelt:
#1661 added. Fixed issue for Sacrificial offering.

Mayegelt:
Power Lances for Command Squads added.

Mayegelt:
Fixed the Dread / Comtemptor / Cortus. Previously by swapping your ranged weapon for a melee one, it was giving you the bonus attack, even if you then swapped the melee one you started with for a ranged one.
Also relinked Meltagun on a few things, as it was once again doubled up for no reason.
Also added a profile to Heavy Bolter with Suspensor Web.

Mayegelt:
Had forgot to add the GST file upgrade for Heavy Bolter in the upload.

PR #1774, opened 2020-09-22
Book Nine

capitaladot:
Take a look and add comment checkboxes like so if we missed stuff. 
- [ ] thing we missed


Mayegelt:
You have put in a Firewing Enigmatus Cable fix for compulsory Scion. They sadly are not compulsory scions for serps bane, they get made into troops, however only seeker and assault get to be Compulsory.

Mayegelt:
I would suggest that the last pull request stuff I've just done is strong enough to put out to the public, unless their are any glaringly obvious issues.
Let people play with the units as as far as I can tell all the points should be correct. Though I've only done the new units, NL units and the Special Chars for DA, as well as the pre-released units., even if the RoW's aren't fully complete. And just tag all the RoW's that aren't 100% finished as (Work in Progress).

Mayegelt:
- [ ] Ironwing robots
- [ ] hiding the Scions of the Hexagrammaton from other legions

PR #1839, opened 2020-12-11
Update (HH) Imperialis Militia and Cults Army List.cat

AsterixCodix:
## Description
Hi there, so first time making a PR for this project (and probably my last time but), it's been an issue that's kinda been bugging me: in the Imperialis Militia and Cults rules, the Advanced Weapons upgraded afforded by Survivors of the Dark Age only affects las weapons and rotor cannons, but the wording of it here implied it affected Bolters too. (If there's an FAQ anywhere that changes that I'm really sorry). It's nothing major so I didn't think I needed to fill out the rest of this but... Tada?


Mayegelt:
What I'm assuming is maybe some hide functions are missing. Or when it was originally typed up, the person did a quick job, rather than a complete one. I'll check the patch soon hopefully.
If I can see any other changes (like hidden values) I'll then throw them in.

PR #1849, opened 2021-01-11
Thousand Sons Cult restrictions added

Mayegelt:
Thousand Sons Cult restrictions added. Did a bit of testing and seemed to be working.
Also fixed an issue with MOTL which was allowing one in lower than it should points values. Also the MOTL restriction is spread over all forces in roster, not each force individually.


Mayegelt:
@capitaladot if you can have a solid play around with this. I've tested it myself, but didn't go beyond no RoW. But the fix should work on all of them.
Interesting note is I worked out you can't use Armoured Breakthrough or Fury of the Ancients in Thousand Sons, as both require a compulsory troop choice that can't have a Cult Arcana.

Mayegelt:
Also did a bit of cleaning up some entries as multiple entries for the same thing were in certain files such as armoured cockpit and extra armour and various weapons

Mayegelt:
Today's Contekar update complete as well
https://www.warhammer-community.com/wp-content/uploads/2021/01/zllrihJyMHFAmO9Z.pdf

Issue #1865, opened 2021-01-20
Maru Skara issue


Mayegelt:
Maru Skara seems to be demanding 3 champions. 1 disappears when taken, the other 2 remain...
Fix implemented by assigning a new "Champion Consol" category with the same limitations of Min in Force 0 -> 1 if you take Maru Skara. No idea why it didn't work, but that fixed it.

Mayegelt:
Sorry this has been delayed. Was a can of worms issue, that found lots of other fixes needed.

PR #1866, opened 2021-01-20
Uniformity of Category -> Entries. RoW Changes & Maru Skara #1865

Mayegelt:
#1865 In Maru Skara RoW for no apparant reason the Champion was being asked for 3 separate times as a force requirement. But taking 1 only seemed to remove the requirement once. Deleting the requirement inside the Consol removed all instances of it asking for the change. Adding it again seemed to trigger the same issue.
Fix implemented by adding a new Category with exactly the same requirements and adding that Category to the Champion Consol. No idea why it wasn't working, I assume some strange lookup issue of a requirement inside of an entry, inside of a group entry, inside another entry, inside an entry...

Going to leave this here for a while unpulled so that if any other issues arise in the next X number of hours I can put them in at the same time.


Mayegelt:
Hammer of Olympia Extra Armour cost
Contemptor Cortus Dread
Damocles Command Tank
Mortis Dreadnought

Siege Breaker or Warsmith issue resolved. Was flagging an error for both needing a 1 Siege Breaker and 1 Warsmith until either was selected. This has been changed by adding a “Siege Breaker or Warsmith” entry with a min/max or 1/1 in each entry, which will aid in clarity.
Also did the same with the recently fixed Champion one for Maru Skara, as Category entries will take up more resource that an entry link buried in 2 entries.
This will affect older lists, but only in regards of them having to click a buried link, which should be findable by just clicking the error.

Sagyar Mazan fix for Enforcing more infantry than vehicles

Mobility options needed to be changed for Praetors and Consols, as it was impossible to fix Sagyar Mazan and Iron Hands without doing so.

Blackshields Reaver Lord boarding shield was incorrect cost.

Fix for Command Squads having Jump, Bike or Jetbikes.

Gave more uniformity to the categories on lots of units. Many were missing Legion Astartes, Character, Independent Character… In addition the Tank, Vehicle and Walker entries were split up between higher and lower levels of the entries, which caused the restrictions of the Dark Angels RoW for Ironwing and Steel Fist are both messed up, hopefully should be fixed.

Mayegelt:
Just put on some more fixes for it.

Mobilty fix for Centurion and Praetor to have default selection
Sicaran Venator fixed
Fixes to a few more units that had Upgrade rather than Model for models
Whirlwind Scorpius fix
Had to remove some category tags inside of the dedicated transport part of units.

This looks all good to me now. Only one that might not be is the old Ironwing Protocol, which I'm unsure how to take the description of how it works...
Dark Angels Ironwing Protocol RoW (At least half the units in the army must be vehicles with the tank type)
Currently if I had a Delegatus & 3 tactical squads, that equals 4 units so if 2 of those units take rhinos then its is ok, as 50% of the units are vehicles of the tank type, as the rhinos are part of the tactical squad unit.
However if you were to buy non-dedicated transports, such as the same 1 delegatus & 3 tactical squads, you would need to take 2 tanks to counter that, + 1 extra per 2 of any additional units for instance 2 spartan (2Unit 1Tanks), 1 whirlwind scorpius (1Unit 1Tanks) and a sabre (1Unit 1Tanks) (making 8 total units and 4 tanks). OR you could take a unit of 2 sabres (1Unit 2Tanks) and a Whirlwind Scorp (1Unit 1Tanks) (making 6 total units and 3 tanks)
Which I think is correct.

Mayegelt:
Ironwing Protocol kind of fixed.
Still no way to enforce the min number of transport capacity to mean all units MUST start in transports, due to things like Spartan being able to hold a squad + apoth + characters...
However, fixed the need 1 tank per 2 units thing by adding in a category to all units at the root entry "Ironwing Protocol Min 50% Tanks Requirement" Though I can probably have a more snappy name. Also this might be a useful thing to have added to all units. So instead of using "Unit" from the Unit/Upgrade/Model selection, it would be a true count of all units, including dedicated transports, which if taken as a sub entry of an existing unit, do not seem to flag as a unit themselves in cases.
HOWEVER this has meant that in the case of the Terrax, Spartan, Stormeagle, Rhino, Proteus, Phobos have all had to be copied to make a unique dedicated transport version, and a non-dedicated transport version. Which could cause some additional work when making amendments.

Mayegelt:
For some time it looks like this was added wrongly to the Legiones Astartes part of the resections rather than the Infantry part. This was then counting non infantry units such as Jetbikes and Bikes as part of the restriction mitigation, which should not have happened.

Mayegelt:
Now it should all be done, including Sagyar Mazan.

Mayegelt:
est Rosters - Descriptions are in the Custom name of the list, but also listed below

BLACKSHIELDS
A Blackshield detachment may never have more units with the vehicle type than it has units with the infantry type.
3 Infantry units are in the army (Reaver Lord, Mauauder, Breacher) The dedicated transports do not count
The Sabre, Arquitors, Leviathans and Preditors are all Vehicle Units
Removing any one of the Vehicle units will resolve the issue,
OR
Adding any other Infantry unit.

BLOOD ANGELS
With the exception of Dedicated Transports, a Blood Angels Detachment may not have more units with the Vehicle type than it has units with the Legiones Astartes (Blood Angels) special rule.
3 Legion Astartes units are in the army (Preator, Breacher, Tact) The dedicated transports do not count
The Sabre, Achiles, Arquitors or Leviathans are all Vehicle Units
Removing any one of these units will resolve the issue,
OR
Adding any other Legion Astartes unit.

Hammer of Olympia
Must have a min of 1 Warsmith or Siege Breaker. This can be resolved by:
1) Giving the Centurion Siege Breaker
2) Giving the Praetor Warsmith
3) Taking Kyr Vhalen as he is a Warsmith instead of the Praetor (This was previously an issue)

IRON HANDS
Iron Hands detachment may not have more units with the Legiones Astartes (Iron Hands) special rule in total (including Independent Characters) with the Jump Infantry, Bike or Jetbike types than it does with the Infantry type.
The Infantry units in this army are Praetor, Breacher, Tactical (3 units)
The Jump, Bike, Jetbike units are Assault, Hussar, Jetbike, Outrider.
NOTE: The Assault squad counts as both Infantry for some things as well as Jump. So has a 2 modifier to the number of infantry units, as it counts as infantry itself.
To resolve the error, add 1 Infantry unit, but not a Jump Infantry unit.

IRONWING PROTOCOL
At least half the units in this RoW must vehicles with the tank type.
This list has the new dedicated transports count at part of the unit turned on. So this army has 3 units total (Praetor, Breacher, Tact), as the dedicated transports do not count as their own units, but as part of the unit confer the "Tank" type to the unit they are attached to.
If you change the list to dedicated transports count as their own unit, it will cause an issue as you will then have 5 units (Praetor, Breacher, Tact, Landraider, Rhino) of which the later 2 are all units consisting of tanks. Which is less than the 50% required.

Maru Skara
Min of 2 HQ, of which 1 needs to be a champion. Was previously giving an issue of flagging the champion issue 3 times. Only removing 1 if you had any champions I the army.
Resolution was to add a tickbox inside of Champion with a min/max of 1. This also helped the Mournival stuff.

RAVENGUARD
A Raven Guard detachment may not have more units in total with the Vehicle (Tank) type than it does with the Legiones Astartes (Raven Guard) special rule.
3 Legion Astartes units are in the army (Preator, Breacher, Tact)
The Sabre, Achiles, Arquitors & Predators are all Vehicle Units
Removing any one of these units will resolve the issue,
OR
Adding any other Legion Astartes unit.
If you change that Dedicated Transports count as their own units, it will then require the removal of 3 Tanks Units OR addition of 3 Legion Astartes units

STEEL FIST
At least half the units in this RoW must consist entirely of vehicles with the tank type.
This list has the new dedicated transports count at part of the unit turned on. So this army has 4 units total (Praetor, Breacher, Tact, Preds), as the dedicated transports do not count as their own units. So also the dedicated transports are not a unit entirely consisting of vehicles with the tank type as they have infantry in there.
To remove the error on list you need to add 1 more unit which  is a tank such as unit of predators or sabre. OR if you switch to dedicated transports count as their own unit, it will resolve the issue as you will then have 6 units (Praetor, Breacher, Tact, Landraider, Rhino, Predators) of which the later 3 are all units entirely consisting of tanks.

BERSERKER ASSAULT
Detachments using Berserker Assault may not take more vehicles with the Tank or Flyer type than they have Infantry units in the detachment
This force has 3 Infantry units (Praetor, Assault & 2 Tacts)
Tanks/Flyers 1 unit of 2 Sabres, 1 unit of 3 Predators (5 tanks/flyers)
Adding 1 additional infantry squad will fix error, any flyers or tanks including dedicated transports will increase the problem.

capitaladot (changes requested):
- [ ] Spartan is still missing as a DT for Iron Fist.

Mayegelt:
Which squad is missing the spartan? @capitaladot 

Remember to have access to the two types of land raiders and spartan you need a model with the sign of iron wing in the unit. Additionally 10 9r less models for the land raiders and 15+ models for Spartan.
Destroyers need Phobos added
Seekers need Phobos added

Issue #1881, opened 2021-02-02
Justearin terminators weapons


Tntm1ner:
When I buy justearin terminators in either elite or troop slot and give them anything else than a combi-bolter or combi-weapon option like a pair of lightning claws or a heavy weapon it say i need more selections from exchange combi-bolter if I buy more combi-bolter it obviously gives me the error that I have too many weapon selections

capitaladot:
It works for me if I just type a 5 in the Combi-Bolter space.

Mayegelt:
Reopening this one. Needs a fix. Looks like the issue it's it's not counting correctly. Will try to push out a fix today.

Mayegelt:
Fixed but just need to push

Issue #1880, opened 2021-02-02
The Black Reaving wants three masters of signal


Tntm1ner:
When I use the black reaving rite of war I get three errors that I need a master of signal if I buy on one dissappears but the other two remain

capitaladot:
Is it possible some say Master of the Legion? You can only have one such per 1000 points.

capitaladot:
Nope, totally checks out; totally broken.

Mayegelt:
This will be the same issue as #1865 then by the sounds of it. I will be able to fix in the same way.

Mayegelt:
Same issue was affecting Logos Lectora. Fixed but just need to push

PR #1882, opened 2021-02-02
Mayegelt 1880 1881

Mayegelt:
Master of Signal - Black Reaving and Logos Lectora #1880 and Justarian #1881

Test rosters added
Ultramarine Logos
Issue with force org asking for masters of signal multiple times. If one is added to this roster or a Damocles rhino it is resolved.

SOH Black Reaving and Justarian.
Issue with force org asking for masters of signal multiple times. If one is added to this roster is resolved.
Also fix for Justarian min/max weapon requirements.


Mayegelt:
@capitaladot once again I've used the forks, Should be ready to go live.

Mayegelt:
Ahrimans Command Squad #1883 fixed. They were lacking the ability to take the ML2 Brotherhood of Psykers when he was warlord.
Fixed by adding an additional selection entry to Ahriman then cross linking it to the original "ancestor of" options from the command squad, which were then changed to Force and that selection entry.

capitaladot (approved):

PR #1888, opened 2021-02-17
Laser Destroyer Array and QoL fixes

Mayegelt:
Laser Destroyer Array correctly readded to some LA entries. It seems that several of them had been replaced with "Laser Destroyer" rather than Array (which has an increased range). 
QoL fixes for Talons Custodes units. This will mean almost all weapon options will be reset. But should make lists look better presented and easier to use in future.
QoL fix for Veteran squads as well. Previously on adding them to the list it gave you none of the default starting equipment. Now it does and can be changed as normal.


Mayegelt:
Oh forgot I also added a basic centurion entry to talons.

capitaladot (approved):

Issue #1905, opened 2021-04-25
Zone Mortalis unit size limit incorrectly applied for Astartes units


ntyrrell:
In the Legiones Astartes army list, there is a modifier that sets the maximum number of models in a unit to 15 if the force is using one of the Zone Mortalis Force Organisations.

There are two errors here; 
1. The rules as presented in the latest PDF from FW (https://www.warhammer-community.com/wp-content/uploads/2019/11/HHZone_Mortalis_Rules.pdf) suggest that only a unit with a starting size of 15 is not allowed in ZM. This limitation comes from using the Zone Mortalis Assault Force RoW, which limits units to 15 models (Red book, p. 104).
2. The modifier is incorrect, as it does not account for the sergeant, allowing for units to have 16 models.

The fix would be to drop the maximum unit size to 14 and to apply this modifier if the Zone Mortalis Assault Force RoW.

Thank you for your time reading this issue.

ntyrrell:
Slight amendment to my issue - upon further review, it appears that the word "starting" is referring to the start of the game, rather than the minimum starting strength of the unit.

However, the second issue is still valid, I believe.

Mayegelt:
Thanks for the report. I'll get on it soon.

Mayegelt:
Ok corrected Assault, Breacher, Medusan Immortals, Phalanx Warders, Tactical.
Not sure if there are any others.

Issue #1904, opened 2021-04-06
Talons of emperor bug


Dorn1512:
Hello, i found a mistake at legio custodes sentinel guard squad.
At BS magisterium vexilla changes shield for free, but at armylist it costs 10 pts.

![sentinel](https://user-images.githubusercontent.com/82025950/113686722-3dda1b00-96d0-11eb-92a3-ff724a78fe44.JPG)

Mayegelt:
Thanks for the catch. Will put it in the To-Do pile

Dorn1512:
PS: and please fix a special rule of legio custodes:
Preternatural Skill: Any model with this special rule gains a bonus of +1 to their Initiative in the Assault phase if their Weapon Skill is **equal to or** greater than the majority Weapon Skill of the unit or units they are fighting. (White Dwarf)

It must be:
Preternatural Skill: Any model with this special rule gains a bonus of +1 to their Initiative in the Assault phase if their Weapon Skill is greater than the majority Weapon Skill of the unit or units they are fighting (HH8)

![Preter](https://user-images.githubusercontent.com/82025950/113687618-3404e780-96d1-11eb-9e73-31d0328dbfef.JPG)

Mayegelt:
Looks like that was a change from book 7 to book 8 which was overlooked

Issue #1903, opened 2021-04-06
"Invisible" Weapons Selections


Lord-Falkenrath:
Apologies for my second thread today, but given the widespread nature of this problem I figured I'd put all the offenders in one post, organized by Legion and Unit. The issue itself is selecting a given piece of wargear and the stats/abilities of said wargear not appearing, even though the points cost is appropriately increased and the item name is often (but not always) referenced in the appropriate keywords/wargear section.

Note that this list does not include any Lord of War options, I do not know if any of those options suffer from similar issues. Again, I apologize for potential bad formatting and the lack of tags, I am new here.

**All Legions:**

- [x] Terminator Command Squad: Twin Lightning Claws (but not single lightning claws), Grenade Harness (on Standard Bearer)
- [x] Centurion in Terminator Armor(Any centurion, any pattern): Twin Lightning Claws (but not single claws)
- [x] Forge Lord Centurion: (Consul Box) Master-Crafted Bolter, Plasma Pistol
- [x] Praetor in Terminator Armor(Any pattern): Twin Lightning Claws (but not single claws)
- [ ] Knight-Errant: Librarian and Psychic Mastery Level upgrades do not specify disciplines, Jump Pack does not confer Jump Infantry type and rules.
- [x] Assault Squad: Sergeant options only Hand Flamer and Twin Lightning Claws
- [x] Breacher Siege Squad: Sergeant options only Single Lightning Claw(Exchange Bolter/Bolt Pistol box only), Thunder Hammer
- [x] Reconnaissance Squad: Melta bombs. Sergeant options (Bolt Pistol Box) Chainsword/Combat blade, Power Fist, Single Lightning Claw; (Bolter Box) Chainsword/Combat Blade, Hand Flamer, Power Fist, Single Lightning Claw
- [x] Tactical Squad: Extra Chainswords/Combat Blades grants Chainaxe item to non-World Eaters.
- [x] All Force Weapons: Force Mauls display Force Sword stats, and vice versa.

**Dark Angels:**

- [x] Deathwing Terminator Companions: Twin Lightning Claws (but not single lightning claws)
- [ ] ~Scion of the Ironwing Upgrade: Not missing, but worded incorrectly. As written it has the unit "treat results of crew shaken as crew stunned", when it is meant to be "treat results of crew stunned as crew shaken".~

**Emperor's Children:**

- [x] Palatine Blades: Jump Packs, Sgt Meltabombs.
[...]

PR #1907, opened 2021-04-28
#1902 #1903 #1904 #1905

### LINKED ISSUE #1902: Collection of Issues I found while making lists
I apologize if some of these have been fixed in the latest update, but I have been collating these issues as I found them while making lists so that I could post them all at once instead of separately. I also do not know how to tag on Github, so I apologize if this post isn't tagged or is improperly tagged.

1) Blood Angels Legion: Chapter Master Raldoran counts as 2 Master of the Legion, should only count as 1. Lacks "Compulsory HQ" tag. Cannot be selected as warlord, but should be.

2) Rite of War: Fury of the Ancients. RoW Requirements on BS are 3 Forge Lords and 3 Primus Medicae, should only be 1 of each.

3) Rite of War: Angels Wrath: The RoW as written has no explicit requirement for "fliers per unit of non-fliers", but BS forces one. Perhaps add a "Check infantry transport capacity manually" warning and remove the requirement?

4) Iron Hands Legion: Gorgon Terminator Hammer Bearers have no weapons options, and their weapon choice cannot be accounted for on the main unit weapon selection. Ie, in a unit of 5 terminators you can only select 4 weapon options, with no further options presented on the Hammer Bearer section.

5) Iron Hands Legion: Castrmen Orth cannot be selected as Warlord, but should be.

6) Various Primarchs: Leman Russ, Angron, and Fulgrim lack the Bulky keyword. Perturabo, Horus, and Mortarion lack the Very Bulky keyword.

7) Salamanders Legion: When selecting a heavy flamer for a weapon option on a DT for a Firedrake terminator squad, BS will flag a warning that the squad as 1 too many weapon selections, despite the weapon being on the DT and not the squad itself.

8) Iron Warriors and Imperial Fists Legions: Taking the Primarch's special DT (the Tormentor and Aetos Dios, respectively) is supposed to be possible in 3000+ point lists. However, it still flags an error "[Primarch] has 1 too many selections" unless the list is 3001+ points. This might be me misunderstanding the intent of the wording, however.

Edit: Added issue 8.

Mayegelt:
Various things updated and fixed

#1902 #1903 #1904 #1905 as references of details.
Though mainly a few issues of Battlescribe flagging errors.
Various weapon profiles missing.
A single Talons issue.
Some ZM force issues about max number of people.


Mayegelt:
Hopefully this should fix a bunch of issue.
Also if Windstorm is right that the issue involving "instance of primary category condition" was what was causing the appspot issue, then hopefull this should resolve it as all of them were removed from the LA file (11 entries spread over 6 units). If not it won't cause any issues as it just changed the "Primary Category" one to just look at "Category" instead, which in all cases was the same result.

PR #2069, opened 2022-06-20
Page Numbers and Verified Special Rules It will not Die through Unwieldy

nstephenh:
Added page numbers and verified some special rules (removed P3P if verified).

### DIFF
--- a/2022 - Horus Heresy.gst
+++ b/2022 - Horus Heresy.gst
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <gameSystem id="28d4-bd2e-4858-ece6" name="Horus Heresy (2022)" revision="1" battleScribeVersion="2.03" xmlns="http://www.battlescribe.net/schema/gameSystemSchema">
   <publications>
-    <publication id="e77a-823a-da94-16b9" name="Horus Heresy Rulebook"/>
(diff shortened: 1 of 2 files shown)


nstephenh:
Added thousand sons wargear

Mayegelt:
I'm having a look at the files and not seeing why certain bits are added or what you mean by them. Though that is because I'm probably terrible at Github :)
So for instance with the Outflank one you have added a comment to it. Is it a case that you noticed a mistake and didn't correct it but planned to?
Same with Master-crafted.

For "Termplate weapons" you comment "Is it possible to nest special rules? Can we give things we give this to wall of death and ignores cover automatically?". Firstly the answer is no :( . But putting it as a comment rather than in the pull request or conversation means it might just end up being lost.

Other that that it just looks like BSData randomness with playing with the sabre transport capacity and access points with no change, but reporting one.

Mayegelt:
Just removing the p3p should indicate it's been checked.

At some point I'll also start moving some of the other shared profiles into the right selection entries as well. So for instance you don't need the autocannon as a shared profile external to the selection entry. It was just done initially for convenience while typing in lots of code to make sure it was all done. And not have to open lots of selection entries to check.
Will need to make sure they keep their ID numbers though for the profile as some will be using multiple links to it such as master crafted ones. As I don't want to have to make a regular and master crafted version of every weapon able to be master crafted. Though might make a few of the more common ones.

Mayegelt:
I believe so.
A lot of stuff was done from playtest. Then the advanced previews from reviewers online started coming out with official books.

PR #2098, opened 2022-07-09
Update 2022 - Legiones Astartes.cat

Toreador13:
Deleted and recreated both Assault Squad and Despoiler squad due to amount of changes.

Added Phoenix, Caedre, and Achean weapon groups.

Cerated a new brand with these changes to try again.


Toreador13:
Instead of trying to go back and make all the changes, I deleted the two entries and recreated them from my good file. I had made too many changes to clean up the entries to do that all over again. Sorry as that creates a large DFF and will mess with anyone playing with lists, but the two entries should no be complete barring any bugs found.

PR #2103, opened 2022-07-11
Moving profiles, fixing Cult Arcana and adding special rules

Mayegelt:
Moving profiles, fixing Cult Arcana and adding special rules name modifiers

### DIFF
--- a/2022 - Legiones Astartes.cat
+++ b/2022 - Legiones Astartes.cat
@@ -20091,17 +20091,6 @@ A Perfect Fury: If a model or models with the Skill Unmatched special rule selec
                 <cost name="Pts" typeId="d2ee-04cb-5f8a-2642" value="5.0"/>
               </costs>
             </entryLink>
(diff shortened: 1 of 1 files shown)


Mayegelt:
@nstephenh big old update here. Let me know if I ended up reversing anything in your patch. I shouldn't have as I think it was just adding pts to paired claws and remove the extra one on Tart Sergeants.
Did notice that Ravens Talons might be wrong price on paired though I think as its set to 0. They should be 5pts each more expensive than lightning claws (or 10pts on a pair)

nstephenh:
Looks like you only got half my fix, I also set the points on paired weapons (though I didn't set the raven talons to be 5 pts more).

Mayegelt:
Copied the pts over for the standard points

nstephenh:
Hard to see on my phone, I'll review tonight. Is there a points modifier based on praetor, etc?

Mayegelt:
Yep.
As said. No rush to review. If it's done by tomorrow afternoon then
hopefully I can get to work on next steps.

nstephenh (approved):

PR #2109, opened 2022-07-13
Did a bunch of fixes for Troop choices

Mayegelt:
Looks like options were missing or hidden for no reason. I seem to have fixed them. Also a few values wrong.


Mayegelt:
If you can have a looky and check the values and options that are on the units. I've also set default values for Bolt Pistols and Bolters and stuff on them as well.

Toreador13:
> Looks like options were missing or hidden for no reason. I seem to have fixed them. Also a few values wrong.
Probably left over for my first pass. Didn't see that Legion items were hidden by default in the entries, and a few weren't working.

nstephenh (approved):

PR #2110, opened 2022-07-14
Cataphractii Command Squad and the great Power Weapon SSE

nstephenh:
This contains my first pass at the Cataphractii Command Squad and the Power Weapon SSE which links to a SSEG containing all the legion options.

Note I didn't finish adding all the legion weapons to their individual SSEs.

We'll also want to create character and heavy subtypes as rules again, so we can use them as filters, and ensure the 

Also I fixed terminator Lightning Claw/Raven's Talon points.


nstephenh:
Also to-do, grenade harness.

Mayegelt:
At a glance this looks fine to push just looking at the changes. Though not well enough to load up on computer to check.
If it works for you @nstephenh then feel free to push it and send out a release.
Might be worth also giving those destroyers their jump packs before doing so.

Mayegelt:
Oh and someone mentioned "Scout" squads might also need a points check as well.

Mayegelt (approved):

Issue #2112, opened 2022-07-14
HH Destroyer assault squads dont have jump packs


AstorathTheGrimm:
destroyer assault squads dont have jump packs

**Device and data (please complete the following information):**
 - BattleScribe version [e.g. 2.03.45]
 - Release version [e.g. v1.2.3]

**Additional context**

Toreador13:
Confirmed Destroyer Assault squads do not have the jump pack listed as well as missing their Special Rules. I can update in my patch.

PR #2114, opened 2022-07-15
Bug fix and added units

Toreador13:
Fixed Destroyer Assault Squad Entry
also bug #2112
Added Seekers
Added Dreadclaw Drop Pod
Added Terror Squad


Toreador13:
Newer commit.

PR #2115, opened 2022-07-15
Toreador13    ugh

Toreador13:
Fixed Destroyer Assault Squad Entry
also bug https://github.com/BSData/horus-heresy/issues/2112
Added Seekers
Added Dreadclaw Drop Pod
Added Terror Squad
Changes of Power Weapon grouping for all default units.
Added Grey Slayers and Grey Stalkers


nstephenh (approved):
Looks good at a glance. Sorry it took me so long to accept it.

PR #2121, opened 2022-07-19
Fast Attack Additions

UrsinePatriarch:
**Added/Modified:** 

- Javelin Squadron
- Proteus Land Speeder Squadron
- Sabre Strike Squadron
- Sky-Hunter Squadron
- Storm Eagle Gunship
- Xiphon Interceptor

There are other fast attack options, but it seems others have claimed them and not submitted. I can do them if asked, but will hold off until then.
Also, apologies for the double commit; it's been a while since I've had to do a PR on here and I did some last minute fixes.


nstephenh (changes requested):
This seems like you've replaced the cat with a catz. Please use a regular .cat

UrsinePatriarch:
Been some time since I had to format Git stuff, so I hope this work.

PR #2154, opened 2022-08-01
Inner Circle Knights Cenobium, Inner Circle Knights Cenobium - Order of Broken Claws, Dreadwing Interemptors, Land Raider Proteus Carrier 

revivedtitan:
Added
Inner Circle Knights Cenobium
Inner Circle Knights Cenobium - Order of Broken Claws
Dreadwing Interemptors

Fixed
Land Raider Proteus Carrier - Land Raider Proteus Carrier was not under the Land Raider Proteus Carrier Squadron, I moved it under and adjusted all dedicated transports to point to the shared id.

battlescribe


UrsinePatriarch:
Probably shoulda asked before changing the Proteus Carrier, mate; it's separated for a reason.
It's the only Dedicated Transport in the game that can be taken in a Squadron, so it needs its own separate SSE so it can be referenced in DT SEGs without referencing the Squadron, which messes with constraints.

PR #2170, opened 2022-08-05
Thousand Sons Legion Units and WLT

Arkangelmark5:
Thousand Sons Legion Units and WLT

## Added Units
Magnus the Red
Sekhmet
Contemptor-Orison
Castellax-Achea
Khenetai
Ammitara Occult

## Fixed Units
Ahriman
Amon

##Added WLT
Evoker of Pain
Magister of Prospero
Eidolon of Suffering

##Test Case
image
image
image
image
image
image
image
image


UrsinePatriarch (changes requested):
Please de-nest Master of the Legion as we discussed.
What's the "*" in Amon's Reliquary for? 
The Osiron Magus can only be taken by itself, I believe.
If an Osiron takes a 2nd force blade, do you have a second way of checking its in-built weapon?
Do Khenetai have any other options for their Force Sword slot? If not, it can just be written as an SE "Pair of Achea Force Swords" that includes 2 links in it rather than needing to select it twice for every model added.
Do you have the Khenetai Blademaster's Artificer Armour set to actually change his Save in his profile if selected?
Evoker of Pain seems to be formatted incorrectly on multiple lines; did you copy and paste it from somewhere?

Arkangelmark5:
Fixed formatting with the Evoker of Pain
Fixed Khenetai Blademaster's Artificer Armour to effect his save instead of his strength
Redid the Khenetai force swords as a Pair of Force Swords SE and added the Force Sword rules to it.
Clarified Orison Magus as a upgrade rather then the dreadnought
Moved Orison Dreadnought Combi weapon, so It's no longer nested under the force blade.
Clarified Amons Weapons to have a Melee and Shooting version

nstephenh (approved):

PR #2172, opened 2022-08-06
Emperor's Children Fixes

Medonoop:
## Added Units

Kakophoni Squad
Sun Killer Squad
Rylanor the Unyielding

## Fixed Units

Fulgirm
Captain Saul Tarvitz
Captain Lucius
Palatine Blade Squad
Phoenix Terminator Squad
Vulkan

## Not done yet

Phoenix Warden Consul

## Test Case

image
image
image
image
image

### DIFF
--- a/2022 - Horus Heresy.gst
+++ b/2022 - Horus Heresy.gst
@@ -123,9 +123,14 @@
             <modifier type="set" field="name" value="It Will Not Die (5+)"/>
           </modifiers>
         </infoLink>
-        <infoLink id="2a4b-7961-1ac1-06d9" name="Bulky (X) (P3P)" hidden="false" targetId="676c-7b75-4b6f-9405" type="rule">
+        <infoLink id="2a4b-7961-1ac1-06d9" name="Bulky (X)" hidden="false" targetId="676c-7b75-4b6f-9405" type="rule">
           <modifiers>
             <modifier type="set" field="name" value="Bulky (4)"/>
+            <modifier type="set" field="hidden" value="true">
+              <conditions>
+                <condition field="selections" scope="force" value="1.0" percentValue="false" shared="true" includeChildSelections="true" includeChildForces="false" childId="3edc-a1b9-6dc6-b1ea" type="equalTo"/>
+              </conditions>
+            </modifier>
           </modifiers>
         </infoLink>
         <infoLink id="058b-7554-0d99-1e7a" name="Relentless (P3P)" hidden="false" targetId="7adf-ac9a-5035-522d" type="rule"/>
--- a/2022 - Legiones Astartes.cat
+++ b/2022 - Legiones Astartes.cat
@@ -1571,7 +1571,6 @@ Sire of the Emperor’s Children - All friendly models with the Legiones Astarte
           <infoLinks>
             <infoLink id="ed97-fc70-8fae-0274" name="Relentless" hidden="false" targetId="7adf-ac9a-5035-522d" type="rule"/>
             <infoLink id="129d-c508-d60a-ccd8" name="Independent Character" hidden="false" targetId="c57d-4820-458a-7ab5" type="rule"/>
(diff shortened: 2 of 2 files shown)


2022 - Horus Heresy.gst
```diff
@@ -123,9 +123,14 @@
             <modifier type="set" field="name" value="It Will Not Die (5+)"/>
           </modifiers>
         </infoLink>
-        <infoLink id="2a4b-7961-1ac1-06d9" name="Bulky (X) (P3P)" hidden="false" targetId="676c-7b75-4b6f-9405" type="rule">
+        <infoLink id="2a4b-7961-1ac1-06d9" name="Bulky (X)" hidden="false" targetId="676c-7b75-4b6f-9405" type="rule">
           <modifiers>
             <modifier type="set" field="name" value="Bulky (4)"/>
+            <modifier type="set" field="hidden" value="true">
```
Medonoop, line 129: Added condition to hide this when using EC so Fulgrim can show Bulky (6) by itself

2022 - Legiones Astartes.cat
```diff
@@ -36114,6 +36174,223 @@ A Perfect Fury: If a model or models with the Skill Unmatched special rule selec
...
+        <categoryLink id="5977-1b6a-1f4c-5601" name="Legiones Astartes" hidden="false" targetId="11f2-472f-c1d1-9ae9" primary="false"/>
+      </categoryLinks>
+      <selectionEntries>
+        <selectionEntry id="b6b8-9e05-1fd8-accb" name="Phoenix Champion" hidden="false" collective="false" import="true" type="model">
+          <constraints>
+            <constraint field="selections" scope="parent" value="1.0" percentValue="false" shared="true" includeChildSelections="false" includeChildForces="false" id="8d03-21ed-3ecd-130f" type="max"/>
+            <constraint field="selections" scope="parent" value="1.0" percentValue="false" shared="true" includeChildSelections="false" includeChildForces="false" id="07bc-438e-ddb3-c8fa" type="min"/>
+          </constraints>
+          <profiles>
+            <profile id="c8cf-9df4-1a7f-1522" name="Pheonix Champion" hidden="false" typeId="4bb2-cb95-e6c8-5a21" typeName="Unit">
```
UrsinePatriarch, line 36215: "Pheonix"

UrsinePatriarch (changes requested):
Seems like you have 6 options selected in Palatine Blades but only 5 models; are you counting the Dedicated Transport by accident?

If the Navaetor is exchanging their chainsword/combat blade, shouldn't its profile be in there, as well? Essentially the default option being the chainsword and the others being selections?

Loyalist/Traitor isn't actually needed as a rule since it's mostly for categorizing things which we're already doing in other ways.
It's just "Set Hidden = false if selection in force =1 of X/Y"

2022 - Legiones Astartes.cat
Medonoop: Fixed typo

2022 - Horus Heresy.gst
nstephenh: I believe we tend to use "set name to bulky (6)" for this type of mod.

Medonoop:
Fixed Palatine Blades counting the transport for weapon counts. 

Moved Novaetor weapon options onto it. The Novaetor can swap the chainsword or the combat blade for a melee weapon, but no one in their right mind would swap the chainsword instead of a combat blade so the chainsword should always show for them.

Removed any Traitor/Loyalist rules from EC units.

Medonoop:
Added Retinue options to Fulgrim.

Fixed max squad size for Phoenix Terminators when they take a standard bearer.

2022 - Horus Heresy.gst
Medonoop: That is on Fulgrim's profile. This is to hide the original Bulky (4) which is printed off for all Primarchs. Otherwise you get a Bulky (4) and Bulky (6) on Fulgrim's rules.

UrsinePatriarch (approved):

PR #2200, opened 2022-08-15
Making Legions into Legions

Mayegelt:
Set the "Legions" entry to not import in the individual catalogues and the min/max set itself to 0 when they are loaded.
Added root entry to the legion with a min/max 1 to the legion catalogues.


UrsinePatriarch (approved):
All set on this end; need to see the actual code in the editor, but you say it works, so I'll have some trust.

Issue #2203, opened 2022-08-17
[Anon] Bug report: 2022 - Horus Heresy.gstz


BSDataAnon:
**File:** 2022 - Horus Heresy.gstz

**BattleScribe version:** 2.03.25

**Platform:** Android

**Dropbox:** No

**Description:** The price of Seeker Squad is 105 points, not 120 as in data file

Issue #2202, opened 2022-08-17
[Anon] Bug report: 2022 - Legiones Astartes.catz


BSDataAnon:
**File:** 2022 - Legiones Astartes.catz

**BattleScribe version:** 2.03.25

**Platform:** Android

**Dropbox:** No

**Description:** Bug: the Phosphex bomb on the Destroyer squads doesn't have a point cost. Should be 10 points

Issue #2186, opened 2022-08-09
Iron Warriors - Graviton Weapons not limited to ICs.


parkerPWNT:
**Describe the bug**
Graviton weapons should only be available to ICs. 
Currently an entire terminator squad could take Graviton Crushers or a vet squad members could take Graviton Maces.

**To Reproduce**
Set legion to IW
Add a terminator squad. 
View weapon options.

**Expected behavior**
Graviton melee weapons should only be available to sergeants or other ICs. 

**Device and data (please complete the following information):**
 Windows 
BattleScribe v2.03.21.166
Horus Heresy (2022) v7

UrsinePatriarch:
Independent Characters only is incorrect (Sgts are not ICs), but Characters only is correct; this will be addressed. This'll be a pain, but fairly workable with more constraints.

Thank you for bringing this to my attention.

UrsinePatriarch:
Still working to mark everyone relevant as character/IC, but Graviton weapons now properly check for those things, so this is as "closed" as it will be for some time.

PR #2210, opened 2022-08-18
Various Catalog Works

UrsinePatriarch:
- Made the main LA catalog a library
- Added Deathshroud back to main catalog for Retinue usage
- Finished Mortarion
- Mortarion has the same problem as Fulgrim in that he's bulkier, so he was added as an exception in the GST
- Added the rest of the World Eaters units
- Made IW Graviton Weapons C/IC only; no tags for them yet, but we'll get there. For now, they're all hidden by default.
- Added Deathstorm Drop Pod Squadron
- Fixed Contemptor Fist mistake, Delegatus conflict
Address #2203 #2202 #2186 (partially; fixes the logic but all the units now need to have that category added)
image


Mayegelt:
Why did deathshroud have yo go in main catalogue?

UrsinePatriarch:
Apologies for all the commits; was trying to correct errors in my previous merge.
Seems to be all of them, though, hopefully.

PR #2224, opened 2022-08-20
Moved Tson Units into it's own cat

Arkangelmark5:
Migrated Thousand Sons to it's own cat
Added Cult Arcana to the Praetor and Centurions


UrsinePatriarch:
Just a heads up; you also added a whole bunch of Cult Arcana stuff that wasn't noted in your update comments; when I had to manually re-update your LA file cause it downloaded corrupted for me, I almost missed them and deleted all of that work.
Thankfully I'm a bit more thorough and paranoid, but please be more specific in the future.

UrsinePatriarch:
Worked with Arkangelmark5 on this in #2235 ; fixed all conflicts and added even more Arcana-y goodness.

PR #2231, opened 2022-08-21
Toreador13   nl   boo  galoo

Toreador13:
WS boogaloo!

Split out White Scar units and completed them including Exemplary and Legacy. Finished Jaghatai. 

Removed units from LA file. 

Found some of the basic equipment on Destroyers was not there, added it.


UrsinePatriarch (changes requested):
Would request that any Warlord Traits in the WS catalog are changed to the WLT profile rather than rules.
What's going on with the Angel's Tears changes over around 9196? Is that just BS goin odd cause you deleted things and it replaced them? Just checking to be sure.
Can't say all of it is 100% correct, but I'll trust you on this and make edits later if anything turns out to be wrong.

alphalas:
The tears shit is probably because I deleted it in mine and that one's already been merged

Toreador13:
The one Warlord trait changed to profile. 

I didn't touch anything with Blood Angels.

UrsinePatriarch (approved):

PR #2245, opened 2022-08-22
RG Complete

Toreador13:
Finished RavenGuard

Strike Captain Alvarex Maun
Moritat-Prime Kaedes Nex

Added rad Grenades to Destroyers in LA as they were missing.


UrsinePatriarch (approved):
Even simpler than usual cause nothing was removed from LA: great work.

PR #2246, opened 2022-08-22
Hid them

Toreador13:
Changed Legion units to hidden if not correct allegiance.

Changed RS entries to not import


UrsinePatriarch (approved):
With the newest commit, it all seems good; amazing, mission complete; that right there is why you're the best, [Mr. Dark].

PR #2248, opened 2022-08-23
Toreador13   nl   boo  galoo

Toreador13:
Split SW out of LA file. 
Added missing units like Leman Russ (?)

Added Chieftains to SoH file.


Toreador13:
Looks like LA file has conflicts due to BA stuff again? LA file should be easy, just remove the SW stuff.

UrsinePatriarch:
Yeah, we're adding so much stuff to and from and adding more stuff in etc etc that it's just a lot overall.
I'll take a look later today.

PR #2251, opened 2022-08-23
Hopefully Merging works here

UrsinePatriarch:
An updated re-pub of #2248; bit tired of having to do it this way, but we're almost all split out, so hopefully this is the last.


Toreador13 (approved):
I think this looks good. Coped down and viewed a roster. One thing we need to change though (which I can do after) is the Jorlund Hunter pack missed my last update. They aren't set to points. 15 each with 125 being unit cost.

PR #2257, opened 2022-08-25
Test

Mayegelt:
Just a test to check for differences.


Mayegelt:
Well I've fixed the Deathsworn thing and put in the BA fix I did for setting allegiance. But everything is loading for me with no errors. So feel free to import this into your own and check again.

PR #2267, opened 2022-08-26
Toreador13   so h

Toreador13:
Sons of Horus split out into their own cat, all units complete.


UrsinePatriarch (approved):
Merged some things and added in all of the relevant wargear; pretty quick and easy this time around. Thanks for all of the hard work, mate ~

PR #2283, opened 2022-08-29
Warlord Traits and more Mech Madness

Mayegelt:
All Legion and the Mech Warlord traits added / finished

--

Not sure if we have any pattern for putting in advanced reactions. I put the Mech one as a rule under the Mech allegiance.


UrsinePatriarch:
Would it not make sense to make the Advancdd Reaction its own profile and nest it under Allegiance?

PR #2293, opened 2022-08-31
Minor Updates to RG and SW

Toreador13:
Missing dedicated transport options on Varagyr and Mor Deythan.
Missing infravisor on RG squads.


UrsinePatriarch (approved):
A lot of the Space Wolves stuff seemed unfinished/unpolished, so I went through the stuff that immediately jumped out at me and re-formatted it.
Otherwise, seems fine overall; we're getting closer to complete catalogs, so minor changes like this are really important.

PR #2319, opened 2022-09-05
RItes of War text and selections

Mayegelt:
Added all the RoW for all universal and legions. Set hidden for legions, though haven't set the hiddens for traitors or anything else like that. Also added the 2 unique Warlord Traits as well.
Nothing mechanically added to files to implement the RoW in any way or put restrictions on armies when they take them.


UrsinePatriarch (approved):
👍

PR #2372, opened 2022-09-11
Bugfixes and Destroyerfication

UrsinePatriarch:
Please see Commits for list of changes.


Mayegelt:
Basic Night Raptors are being charged 10pts for the chainsword, the rest looks ok.
I'd also make the default option for the 1-5 as the hand flamer as its the cheapest options, rather than the more expensive flamer.

Issue #2411, opened 2022-09-16
[Anon] Bug report: 2022 - Horus Heresy.gstz


BSDataAnon:
**File:** 2022 - Horus Heresy.gstz

**BattleScribe version:** 2.03.25

**Platform:** Android

**Dropbox:** No

**Description:** Magnus the red currently has no points cost, it simply reads no cost

Issue #2410, opened 2022-09-16
LA - Delegatus Retinue


Arkangelmark5:
Centurion in Tartaros armor with Delegatus upgrade is missing its retinue.

BattleScribe Mobile
v16. 1.8

UrsinePatriarch:
![image](https://user-images.githubusercontent.com/46959832/190617148-fb03d347-3d8c-4259-8394-16ddfd0c3805.png)
Doesn't seem like it, mate. You sure?

Arkangelmark5:
It's not on mine 
![image](https://user-images.githubusercontent.com/53609638/190651201-69abcd3e-92e4-46c3-9fb7-15d5a7d6cead.png)

PR #2412, opened 2022-09-16
Thousand Sons Bug fixes and new Exemplary Battle unit

Arkangelmark5:
## Added Units
 Numerologist Cabal from latest Exemplary Battle
Added Fortress of the Mind Reaction
## Fixed Units
Fixed Magnus Points Cost
Fixed Character Tag on Ammitara
### Related Issues

https://github.com/BSData/horus-heresy/issues/2411#issue-1375327641
https://github.com/BSData/horus-heresy/issues/2410#issue-1375321418

### DIFF
--- a/2022 - LA - Thousand Sons.cat
+++ b/2022 - LA - Thousand Sons.cat
@@ -255,6 +255,15 @@
       </categoryLinks>
       <entryLinks>
         <entryLink id="7e40-1a85-bbea-36c2" name="Warlord Traits" hidden="false" collective="false" import="true" targetId="631b-b029-0909-b8f6" type="selectionEntryGroup"/>
+        <entryLink id="9876-987e-6f65-d685" name="Retinue" hidden="false" collective="false" import="true" targetId="39e7-5296-42e0-fe94" type="selectionEntryGroup">
+          <modifiers>
+            <modifier type="set" field="hidden" value="true">
+              <conditions>
+                <condition field="selections" scope="parent" value="1.0" percentValue="false" shared="true" includeChildSelections="true" includeChildForces="false" childId="35d4-0dcd-f7d2-a6da" type="lessThan"/>
+              </conditions>
+            </modifier>
+          </modifiers>
+        </entryLink>
       </entryLinks>
     </entryLink>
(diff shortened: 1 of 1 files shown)


UrsinePatriarch:
Took a quick look and it seems like the Tartaros Centurion specifically didn't get Retinue posted in its Root while the Cata Cent got it posted under its WLTs; did you fix both? On phone rn so I can't check.

Arkangelmark5:
Cataphractii Centurion had the Retinue in both the root and WLT, removed the one in the WLT. Looks like it just got copied wrong and that's why the Tartaros was missing it's and the Cataphractii had 2

2022 - LA - Thousand Sons.cat
```diff
@@ -2653,6 +2665,358 @@
...
+              </constraints>
+              <entryLinks>
+                <entryLink id="c982-8a52-79bf-ed69" name="Plasma Gun" hidden="false" collective="false" import="true" targetId="f0d1-332b-c719-ede7" type="selectionEntry">
+                  <costs>
+                    <cost name="Pts" typeId="d2ee-04cb-5f8a-2642" value="15.0"/>
+                  </costs>
+                </entryLink>
+                <entryLink id="4be4-fb04-7db1-9408" name="Bolter" hidden="false" collective="false" import="true" targetId="1ade-0c02-5612-252b" type="selectionEntry">
+                  <modifiers>
+                    <modifier type="set" field="name" value="Master-crafted Bolter"/>
```
nstephenh, line 2745: Set name to master-crafted doesn't work, and the points are set to 10 instead of 5

nstephenh (changes requested):
The bolter is wrong, and needs the option to take the ts variation of the bolter. I also think we need to add the publication, we might be able to add the publication just in this file instead of in the main files.

2022 - LA - Thousand Sons.cat
nstephenh: I have since been informed that "set name to master-crafted bolter" does work on the output. I still don't like it, but it can slide. Weapon should still be 5 points though.

nstephenh:
Additionally, doesn't hide when exemplary units are turned off.

nstephenh:
Fixed those on my own branch, pulled the changes in.

PR #2454, opened 2022-09-27
Toreador13   space wolves

Toreador13:
Slayer and Stalker updates - SW
Collective power weapon options - LA
Collective Combat Shield - LA


UrsinePatriarch (approved):
Not very thorough work, unfortunately, but easy enough for me to fix.
Overall a good production, but many minor mistakes and odd formatting choices.

PR #2550, opened 2022-11-03
Apply Rites of War from Blood Angels to all Legions that aren't Dark Angels

nstephenh:
This branch copies Blood Angels to LA Template.cattemplate,
and then those changes to all the individual legions using https://github.com/nstephenh/BSCopy

See readme.md for how BSCopy works.

This isn't all the ROW but it's all of them that we have so far.

Note:
The BSCopy output does mess up white space a bit. If it would help I can just save each file in battlescribe to return the whitespace to normal.


UrsinePatriarch (approved):
May God have mercy on us and this commit; I ship this merge with hopes for a brighter future for us o7

PR #2643, opened 2022-12-09
Sw row toreador

Toreador13:
SW RoW complete.

Night Lords RoW under way. Need tweaks for both swift blade and Terror Assault. But mostly done.

RG complete except figuring out method for adding allied detachment checks to liberation force

category tag updates, antigrav added.


Mayegelt:
Is that a request for help on a way to do the remaining things, or just saying they are left to do?

Toreador13:
Just what's left to do on those RoW. Those bits are going to take a lot longer than I have today, so more so everyone knows what's left here to do .

Mayegelt:
Cool SA one is nice and easy as I've done it in the past

PR #2645, opened 2022-12-10
RG Libraration force update

Mayegelt:
RG Libraration force update, requires my Solar stuff from my push to work.

Added in GST

Solar Auxilia category with a min 0 in roster and made to min 1 if Liberation Force in Roster
Added that category to the Solar Auxilia SE

**Allied detachment**
max 1 selections in force
modifier to put max 1 in force to max 0 if Liberation Force in Force.
min 0 selections in roster
modifier to put min 0 iin roster to min 1 if Liberation Force in Force.

**Solar Auxillia**
min 0 in roster
Set min 0 in roster to 1 if Liberation Force in Roster.

ADDED **Liberation Force Allies Restriction** SE
Min 0 in parent
Max 1 in parent
Set min 0 in roster to 4 if Liberation Force in Roster.

ADDED the new Liberation Force Allies Restriction SE to any unit in SA (on my files)


Mayegelt:
image

Mayegelt:
Looks like i need to do more work. It seemed to work then didn't...

Issue #2650, opened 2022-12-14
[Anon] Bug report: 2022 - LA - Thousand Sons.catz


BSDataAnon:
**File:** 2022 - LA - Thousand Sons.catz

**BattleScribe version:** 2.03.25

**Platform:** Android

**Dropbox:** No

**Description:** Vigilator has not free nemesis bolter and melta bomb
Pathfinder has not free astartes shotgun.

Mayegelt:
Fixed in release [v16.4.0](https://github.com/BSData/horus-heresy/releases/tag/v16.4.0)

PR #2658, opened 2022-12-22
RoW update for Tsons

Arkangelmark5:
The Guard Of The Crimson King
Added restrictions for Allied Detachments
Added Sehkmet to troops
Added constraints for Magnus, Amon and Ahriman

The Achaean Configuration
Added constraints for Techmarine Covenant and Praevian
Added Castellax-Achea to Troops

## Not done yet
Constraint for the various praetors as the Astartes Legion file can't share information from the Tsons Cat.
One workaround is to move Magnus, Amon and Ahriman to the main Astartes Legion Cat

### DIFF
--- a/2022 - Horus Heresy.gst
+++ b/2022 - Horus Heresy.gst
@@ -895,7 +895,7 @@ In addition, when a Fast Vehicle moves, it may choose to move at Flat-out:</desc
             <constraint field="selections" scope="parent" value="2.0" percentValue="false" shared="true" includeChildSelections="true" includeChildForces="false" id="fa4b-a703-9dbf-bb6a" type="max"/>
           </constraints>
         </categoryLink>
-        <categoryLink id="4240-0870-e7ec-839e" name="Rite of Warzz:" hidden="false" targetId="d494-e450-d4aa-579a" primary="false"/>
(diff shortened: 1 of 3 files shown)


Mayegelt:
Haven't had a good look at the request. Kinda busy today and tomorrow. But seeing your to-do list. Here is my solution. I think it has been done for a couple of other universal RoW.

The Praetor / Magnus / Ahriman constraint can be made inside LA file.
Create a new Category in LA. Then have a modifier in all 5 of the units that adds the category to that unit if that RoW and if they are satisfying the requirement.
Have the category with a min in force of 0. Modifier with set min to 1 if that RoW is in force.

Arkangelmark5:
Updated The Guard Of The Crimson King Compulsory HQs. Everything should be working across both Rites now

Arkangelmark5:
Also fixed issues from https://github.com/BSData/horus-heresy/issues/2650#issue-1495665282, though they aren't technically Tson issues. 
Also added Scout/Infiltrate special rules to the vigilator

Mayegelt:
Had to make a few fixes to the pull request, so redid what was needed, but couldn't make a pull request to your branch/fork for some reason.

1) Currently it is enforcing that every centurion needs to be a praevian instead of just 1 per force.
2) it is also asking for Court of the Crimson King ROW requirements to be fulfilled in every force, not just crimson king. 
3) Castellax-Achea if taken as elites, are requiring more than 1. This should only be for troops.

PR #2659, opened 2022-12-30
Row  - toreador

Toreador13:
RoW for World Eaters and Emperor's Children complete.

### DIFF
--- a/2022 - Horus Heresy.gst
+++ b/2022 - Horus Heresy.gst
@@ -11472,6 +11472,13 @@ Limitations
 • A Detachment using this Rite of War may include only a single Heavy Support choice.</description>
             </rule>
           </rules>
+          <infoLinks>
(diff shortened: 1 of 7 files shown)


Mayegelt:
Id accidently pushed my commit over yours, but then rolled it back again as I was only trying to rename my pull request and it seemed to push it when I did something then.
So I've got a new request to go over yours with patchs.
Because of the TS one being approved it made your LA have conflicts. But can always just close the pull request, and remake as a new upload and it will fix (ideally after pulling my stuff first if you think its good).

Mayegelt:
Looks like the GST might have gone bad with the amount of removals compared to additions. Not sure what is wrong there.

Mayegelt (commented):
Looks like GST has lost a lot of code.

Issue #2660, opened 2022-12-31
Kharybdis unavailable as Leviathan DT


WetworkSicknote:
**Describe the bug**
In the Leviathan Dreadnought Talon entry, a single-Dreadnought Talon should be able to select a Kharybdis Assault Claw as a Dedicated Transport but can't currently. 

**To Reproduce**
(select Leviathan Talon; only DT option is Dreadnought Drop Pod) 

**Expected behavior**
Should see Kharybdis alongside Dreadnought Drop Pod as DT option

**Screenshots**

**Device and data (please complete the following information):**
 - OS: Android 13
 - BattleScribe version v2.03.25.338
 - Release version 16.3.0

**Additional context**

Mayegelt:
Fixed in release [v16.4.0](https://github.com/BSData/horus-heresy/releases/tag/v16.4.0)

PR #2663, opened 2023-01-02
Patches for row   toreador

Mayegelt:
Added the TS changes to your pull request in this one. I don't think i've messed any of your work up. It all looked good my end.

Haven't been able to test the full pull yet.

World Eaters Crimson Path
There were some issues with the infantry / non infantry settings.

It was applying the restrictions to every army 
It also also keeping units tagged as Infantry even if they had dedicated transports.
It was also counting in some instances multiple of something in a unit as multiple instances in a unit of being non-infantry in some dreadnoughts and vehicles.

I have also tagged Assassins, Knights & Titans as Non-Infantry in their root, and Thallax that a Forgelord can take as Infantry in root.
There might be an issue with Fortifications currently that they can't be tagged either way. But either the category can be moved to GST. Or maybe this is the point in which a "Fortifications Library" could be made, then linked to LA and the other libraries. Then the roots added to each rather than imported.


Mayegelt:
Predators and Rampages didn't have a root for FA and Troops respectively.

The rampages had a category add for troops and compulsory troops. This didn't work as it was not primary for troops. But also you can still take them in their normal slots, Not just the new slot.
This just required a new root entry.

Mayegelt:
Added fix for https://github.com/BSData/horus-heresy/issues/2660 Khadbris for Levi's and also fixed WE Berserker Assault Preditors (otherwise it would have doubled up their selections to be both FA and Heavy Support

Toreador13 (approved):

PR #2671, opened 2023-01-13
RoW Updates again

Mayegelt:
SoH RoW complete
Word Beaers RoW complete
Space Wolves RoW Fix


Mayegelt:
Some RG and NL RoW buffs were being applied wrongly, so were giving you their buffs when not having that RoW, rather than when having it.

Emperors Children RoWs

PR #2732, opened 2023-02-26
Updated custodes to remove placeholders and finalize values

th3proj3ct:
## Added Units
None

## Fixed Units
All placeholder Custodes

## Test Case
Screenshot 2023-02-26 163839


Mayegelt:
Thanks for the work on this one. Looked good. Made a tiny mod after to move the statlines to the model level like the Heteron guard had. There was also a broken bit for that unit as well that the Frag Grenades, Armour and something else was counting as a choice. So if you had 10 models it thought you have 14 or something. That was fixes as well.

PR #2806, opened 2023-04-18
Imperialis Militia

Mayegelt:
New Imperialis Militia found here. Started work on it. Hopefully will not take a huge amount of time to get a rough version with placeholders, then work from there.

https://www.warhammer-community.com/wp-content/uploads/2023/04/j2TwD3c2AB8oLsVW.pdf


Mayegelt:
Pts should be correct for units and per model. Need to add in all the options, profiles and other stuff.

PR #2896, opened 2023-07-18
Minor fixes and whitespace

nstephenh:
Just a few minor fixes. 
Not switching to newrecruit at this time, but I did run some of these files through NR which trimmed some excess whitespace.

## Fixed Units
Secutors
Vulkan
Excindo battle automata
Red Hand Destroyer Ravager


Mayegelt:
Will the NR changes revert if changed I'm BDData Editor?
Looking at the mech one it's like trying to find a needle I'm a haystack.
Additionally I noticed that the catalogue version from 26 (current) to 27. Didn't check the other modified files.

nstephenh:
I need to test if the NR changes will revert.
Additionally, per recommendation of @Mad-Spy I will not rename files since BS won't clean up the old ones and then start throwing errors about duplicates. Might try changing the internal names to match the file names though, which should  get around that.

nstephenh:
Battlescribe does still save numbers as 1.0 while NR saves as 1, etc. 
This would at very least be annoying to switch back and forth all the time. 
@Mayegelt are you likely to start using new recruit? I think we should probably switch to it's editor but I can still use the old one for a bit if you want/while we wait for some more features.

Mayegelt:
I'm way to used to the old editor.
A simple fix for yourself if you want to use NR editor is save your file initially in NR, then open it in BDData Editor and then save it again. That I assume will reset the values again to the 1.0s or whatever maybe...

Mad-Spy:
@Mayegelt they are basically the same, the NR one just has a bunch of QoL features that make developing easier. I'd strongly recommend checking it out.

nstephenh:
> I'm way to used to the old editor. A simple fix for yourself if you want to use NR editor is save your file initially in NR, then open it in BDData Editor and then save it again. That I assume will reset the values again to the 1.0s or whatever maybe...

That's what I did. 
I suggest you switch but we should switch at the same time, so delaying the switch.
The few minor fixes are ready for review.

Mayegelt:
> > I'm way to used to the old editor. A simple fix for yourself if you want to use NR editor is save your file initially in NR, then open it in BDData Editor and then save it again. That I assume will reset the values again to the 1.0s or whatever maybe...
> 
> That's what I did. I suggest you switch but we should switch at the same time, so delaying the switch. The few minor fixes are ready for review.

An issue I could see with switching is that other programs that also use the BS data might end up being affected. So I would also ideally not wish to move away to one alternative that could inadvertently cause issues for others programs.

Mayegelt:
I'll try to do a review of the push later this week if I can. Still not very well. But I'm seeing the issues piling up and most should be fixable in a couple of hours work.

Mayegelt:
@nstephenh 
Sorry it's taken a while for me to have a look at this.
So you have removed some whitespace as the title says, but in some cases the whitespace I believe was being used to order the unit entries and options. So with the removal I think some might need 1), 2), 3)... in there to put the unit leader at the top. Though it's not essential.

Also I saw you put a catalogue link in DA to Mech Library. This is not required as the file is already linked to the LA library, which links to the Mech Library. So now it is double linked. It doesn't cause any issues. But it is just unnecessary.

It seems like a few of the pts changes you made from a number with a ".0" or whatever to no longer having that. There are a lot in Dark Angels & Mech and a couple in Salamanders. & World Eaters I'm not sure if you left them in deliberately, or if they will cause any disruption on BS or other programs.

nstephenh:
Remade this as #2905

PR #2945, opened 2023-10-07
Lords of War Overhaul

Mayegelt:
Transferred LoW into their own Library with his own FoC, and removed LoW from the Crusade detachment.


nstephenh:
Two comments on this direction:
1. I think we'll want to have some instructions for lords of war. Perhaps in the crusade list, we should add an entry that says "add a LOW via the force and detachment" 
2. Can we move the "legion" choice from the LA entries to the LOW detachment? Or alternatively, could we drop the LOW force type and keep these entries in the respective forces, just having the separate detachment type?

nstephenh:
Also, there's the uncategorized greater brass scorpion.

nstephenh (changes requested):
See issues in text comments. Also, the numbers have all changed between 0 and 0.0 again, which makes the diff huge.

PR #2996, opened 2023-11-11
Fixed the cargo-8s pintle mounted weapons being added automatically

### LINKED ISSUE #2995: Militia Cargo-8 Transport upgrade options incorrect
**Describe the bug**
The Militia's Cargo-8 Hauler's upgrades are currently incorrect. The armoured container upgrade is listed at 20pts, rather than the 10pts from the army list pdf. This may be because the 2 extra heavy stubbers (each 5pts) you can take with it have been (accidentally) rolled in, but these are in fact optional: _"Additionally, a model with this upgrade may take up to two additional Pintle Mounted heavy
stubbers (these weapons may not be further upgraded)."_

**To Reproduce**
Make a Militia list
Add a Cargo-8 Hauler
Add the Armoured Container upgrade

**Expected behavior**
Armoured Container upgrade should be listed at (and cost) 10pts. And Heavy stubbers should be a variable numbered choice between 1 and 3 (max 3 if armoured container, max 1 if not).

**Device and data (please complete the following information):**
OS: Windows
Battlescribe
2.03.21
Data release version v17.4.1

The4D6:
The armoured container upgrade has now been re-costed to 10pts

The additional stubbers are now added via an option that appears only when you select armoured container, and cap at 2 total (not including the one selectable with or without the container)

image

Fixes #2995


The4D6:
Have had this tested by the person who raised the issue as they're a friend of mine, I'm happy with the implementation, and so are they.

PR #3005, opened 2023-11-15
Added selection to aetheric conduit

### LINKED ISSUE #3004: Ruinstorm Daemons can't select psychic discipline with Aetheric Conduit
After selecting the Aetheric Conduit upgrade, there's no option to select a psychic discipine.

They should be able to select from Biomancy, Pyromancy, Telepathy, or Diabolism

The4D6:
Added radio button selection to the aetheric conduiit upgrade.

As it appears multiple times across the force, this is added to the root selection so it will be available to any unit that selects it.

Diabolism had to be copied in as it's own entry here as it would otherwise reference back to Legions Astartes and create a weird circular dependency.

image

Fixes #3004


The4D6:
Must be PR'd in AFTER #3003

The4D6:
Closing to try and fix CI issue, will open new PR

PR #3007, opened 2023-11-15
Restructured Diabolism and Aetheric Conduit

### LINKED ISSUE #3004: Ruinstorm Daemons can't select psychic discipline with Aetheric Conduit

The4D6:
You can now select psychic disciplines for the Aetheric Conduit upgrade in daemons.

This has been added to the root upgrade item thingy, so any unit that can select it gets to then make their selection from Aetheric Conduit. As such it creates a sub-menu rather like power weapons

Fixes #3004 

image

[This branch and PR have been created to try to fix CI issues]


The4D6:
Issue is fixed, Diabolism has been moved to the core Horus Heresy file, as both Daemons and Marines need to inherit it.

It can't be in marines, because that imports daemons, and daemons would need to import it in turn, and it creates a very strange circular dependency error

PR #3012, opened 2023-11-20
Re-add heavy stubber, fix wounds on Vorax and Arcuitor

nstephenh:
## Fixed Units
Vorax
Arcuitor


The4D6 (approved):
Good catch on the mech units

Issue #3023, opened 2023-11-28
Militia Lieutenant BS incorrect under Legacy of the Great Crusade


The4D6:
Militia Lieutenants in Command Cadres already have BS4 and therefore do not benefit from the +1BS .

Currently it is increasing his BS to BS5, it should not be.

PR #3024, opened 2023-11-28
4d6/militia lotgc fix

The4D6:
- Militia Force Commanders now correctly get Initiative 5 under Legacy of the Great Crusade
- Militia Lieutenants now do _not_ get BS5 under Legacy of the Great Crusade


The4D6:
Fixes #3023

PR #3025, opened 2023-11-28
17.4.4 Release Thread

The4D6:
### 2 Issues Fixed
- #3023 Militia Lieutenants no longer get BS5 under Legacy of the Great Crusade
- Militia Force Commanders now get Initiative 5 under Legacy of the Great Crusade


The4D6:
Fixes #3023

The4D6:
Issue found and fixed:

IW Dominator Cohort was only set to un-hide if exactly one selection of "Exemplary Units On" was present. so if you had it enabled in an Allied Detachment, then they would hide themselves if it was also enabled in your Primary.

This has been changed so they become visible if there are greater than zero selections of "Exemplary Units On" in the roster.

The4D6:
Issue found and fixed:

Skyreaper Battery on the defence line is max 1 in the book, was previously uncapped.

It is now set to min 0, max 1.

The4D6:
New Imperialis Militia provenances of war have now been completed and are ready on #3028 

#3028 has the latest catalogue numbers on it, so it needs to be merged in after #3026

PR #3053, opened 2023-12-28
Number force entries to ensure users find the crusade force

nstephenh:
This probably makes output look slightly worse, but it ensures that the default entry selected is "Crusade Force Organization Chart" or the specific force org chart for questoris knights/titans

Renamed Crusade Force Organisation Chart to "1. Crusade Force Organisation Chart", etc. Zone Mortalis now starts with ZM. Entries that override the crusade chart now start with "0. " so they pop up first.


nstephenh:
After some discussion of if we could do this with a zero width space, we tried that and it didn't work, so I'll add this to the v18 release branch.

PR #3067, opened 2024-01-08
Fixed Red Hand Destroyers "Collective" Weapons

The4D6:
Red Hand Destroyers now have their 1-in-5 weapons split in to two model selections (hammer or rad missile launcher).

I could not find for the life of me where collective was ticked (if it was) for these guys, so splitting in to two made more sense.

They dynamically adjust their max value in the unit based on the unit size but also based on how many of the other 1-in-5 selection there are.

This one requires feedback.

### DIFF
--- a/2022 - LA - World Eaters.cat
+++ b/2022 - LA - World Eaters.cat
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<catalogue xmlns="http://www.battlescribe.net/schema/catalogueSchema" id="d901-0eca-48b5-304a" name="LA -  XII: World Eaters" revision="36" battleScribeVersion="2.03" authorName="UrsinePatriarch" authorContact="BSData Developers Discord" library="false" gameSystemId="28d4-bd2e-4858-ece6" gameSystemRevision="72" type="catalogue">
+<catalogue xmlns="http://www.battlescribe.net/schema/catalogueSchema" id="d901-0eca-48b5-304a" name="LA -  XII: World Eaters" revision="37" battleScribeVersion="2.03" authorName="UrsinePatriarch" authorContact="BSData Developers Discord" library="false" gameSystemId="28d4-bd2e-4858-ece6" gameSystemRevision="72" type="catalogue">
   <entryLinks>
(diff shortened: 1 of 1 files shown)


nstephenh:
This looks like a Battlescribe only bug, I wonder if anything else is affected in the same way. It would theoretically be anything with a SEG (not SEG link) in a model that can have multiple options. 

The other solution I could think of would be to move the SEG out of the unit and make it a link. Or don't fix it, and tell people to use NR instead.

nstephenh:
Ok, so after talking more to other devs and fully understanding the issue, the issue is that collective is set on the linked thunder hammer and the linked missile launcher, so they all try and take the same option. So there are two paths we can go down here:
1. Link them to non-collective entries (make the missile launcher non-collective)
2. Split them into two collective options. 

I think 2. works better from a collection standpoint EXCEPT they can also upgrade their chainswords to a melee weapon. So, I think you need to do 3 options

- Missile launcher and chain sword (1 in 5)
- Thunder hammer and chain sword (1 in 5)
- Heavy weapon and sword selection (1 in 5)

The 3rd option will certainly be taken the least, but it's the only way to get the "collective" unit we want.

2022 - LA - World Eaters.cat
```diff
@@ -3255,19 +3255,24 @@
                 <cost name="Pts" typeId="d2ee-04cb-5f8a-2642" value="15"/>
               </costs>
             </selectionEntry>
-            <selectionEntry id="34ff-115f-b2ad-9c40" name="Ravager w/ Heavy Weapon (1 in every 5)" hidden="false" collective="false" import="true" type="model">
+            <selectionEntry id="34ff-115f-b2ad-9c40" name="Ravager w/ Rad Missile Launcher (1 in every 5)" hidden="false" collective="false" import="true" type="model">
```
nstephenh, line 3258: Because these guys melee choice isn't collective they don't group as collective. Also, their "two blades" option is collective so that causes some weirdness. 

Finally, should we regen the ID on the heavy weapon guy that turned into rad missile launcher so someone who had a list with these before doesn't find their thunder hammers all turned into missile launchers?

nstephenh (changes requested):
See non-review comment.

PR #3072, opened 2024-01-11
Replace non-breaking space with regular space.

nstephenh:
We had a large number of non-breaking space characters that don't really do anything positive for output and were copied from the source text. I've removed them.


2022 - LA - Sons of Horus.cat
```diff
@@ -4643,7 +4643,7 @@
       </constraints>
       <rules>
         <rule id="cc22-87b0-6a3b-bbb6" name="Born Survivor" hidden="false">
-          <description>The first time in any game when a model with this special rule is reduced to 0 Wounds for any reason or otherwise removed from play as a casualty, the controlling player must immediately make a Leadership test for that model. If the test is failed then the model is removed as a casualty as normal, but if the test is passed then the Warlord is not removed as a casualty, remains in play and regains 1D3 Wounds.</description>
+          <description>The first time in any game when a model with this special rule is reduced to 0 Wounds for any reason or otherwise removed from play as a casualty, the controlling player must immediately make a Leadership test for that model. If the test is failed then the model is removed as a casualty as normal, but if the test is passed then the Warlord is not removed as a casualty, remains in play and regains 1D3 Wounds.</description>
```
The4D6, line 4646:
```suggestion
          <description>The first time in any game when a model with this special rule is reduced to 0 Wounds for any reason or otherwise removed from play as a casualty, the controlling player must immediately make a Leadership test for that model. If the test is failed then the model is removed as a casualty as normal, but if the test is passed then the Warlord is not removed as a casualty, remains in play and regains D3 Wounds.</description>
```
D3 is the official templating not 1D3

2022 - LA - Sons of Horus.cat
```diff
@@ -6345,7 +6345,7 @@ A Warlord with this Trait gains a bonus of +1 to its Leadership Characteristic w
...
+                    <characteristic name="Text" typeId="c68e-2cda-b67b-baca">When a Warlord with this Trait is engaged in a Challenge with an enemy model that has a Weapon Skill that is equal to or greater than the Weapon Skill of the Warlord with this Trait, then the Warlord with this Trait gains +1 Strength and +1 Attacks for the duration of that Challenge (if both models in the Challenge have this Trait, then both models gain the bonus regardless of the Weapon Skill of either model). Furthermore, if the Warlord with this Trait causes an enemy model with a Weapon Skill equal to or greater than the Weapon Skill of the Warlord with this Trait to be removed as a casualty during a Challenge, the controlling player of the Warlord with this Trait gains 1 Victory Point in addition to any gained from any other rule or mission objective. In addition, an army whose Warlord has this Trait may make an additional Reaction during the opposing player’s Movement phase as long as the Warlord has not been removed as a casualty.</characteristic>
```
The4D6, line 6348: There is something off with this tag but I don't know what it is. Doesn't seem to be recognised as a proper tag.

The4D6 (changes requested):
Two small changes, otherwise looks good to me

2022 - LA - Sons of Horus.cat
nstephenh: But that's just githubs's formatting, not anything to do with this PR or the data before/after, right?

PR #3074, opened 2024-01-12
Fixed Vigil Storm Shield Issue

The4D6:
The terminator praetors use the "terminator ranged weapon" selection group for their ranged weapon.

Added a hiding condition to the vigil stormshield if the ancestor is a tartaros or cataphractii praetor.


The4D6:
Also fixed templars rules naming issue

PR #3075, opened 2024-01-13
Tidied Up Tech-Priest Auxilia

The4D6:
Ephemera incursus is now hidden.

Their battlesmith now has the proper value in it,

### DIFF
--- a/2022 - Horus Heresy.gst
+++ b/2022 - Horus Heresy.gst
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<gameSystem xmlns="http://www.battlescribe.net/schema/gameSystemSchema" id="28d4-bd2e-4858-ece6" name="(HH V2) Horus Heresy (2022)" revision="97" battleScribeVersion="2.03" type="gameSystem">
+<gameSystem xmlns="http://www.battlescribe.net/schema/gameSystemSchema" id="28d4-bd2e-4858-ece6" name="(HH V2) Horus Heresy (2022)" revision="98" battleScribeVersion="2.03" type="gameSystem">
   <publications>
     <publication name="Github" hidden="false" id="e2a4-ac85-1bef-22f5" publisherUrl="https://github.com/BSData/horus-heresy" shortName="BSData/horus-heresy"/>
     <publication id="e77a-823a-da94-16b9" name="Warhammer: The Horus Heresy - Age of Darkness Rulebook" shortName="Main Rules" publicationDate="June 2022"/>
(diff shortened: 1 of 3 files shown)


2022 - Mechanicum.cat
```diff
@@ -1830,6 +1830,7 @@
     <selectionEntry id="0eae-0c52-1087-a3b0" name="Tech-Priest Auxilia" publicationId="bde1-6db1-163b-3b76" page="28" hidden="false" collective="false" import="true" type="unit">
       <categoryLinks>
         <categoryLink id="0f49-7308-c6f1-e967" name="Line Sub-type:" hidden="false" targetId="6399-5c65-7833-1025" primary="false"/>
+        <categoryLink targetId="86dd-51d8-3669-68ca" id="2b8b-9a5c-efda-9e8b" primary="false" name="Tech-Priest Auxilia"/>
```
nstephenh, line 1833: No idea if these render anywhere but do we want to consider setting this to hidden?

nstephenh (commented):
One minor question.

nstephenh:
This is my ignorance speaking, but is category the best way to handle this? seems like it creates a lot of categories. Guess that might be necessary though.

The4D6:
New recruit allows for a vertical reference to the Magos Auxilia but the battlescribe data editor doesn't support it.

This works with both

nstephenh:
If this isn't set to hidden it'll render on list outputs right? Should we set it to hidden? Or will it not show up?

nstephenh:
I don't know if this is pre-existing or not, but servo-arms don't appear to be selected by default:
image

2022 - Horus Heresy.gst
```diff
@@ -1306,6 +1306,7 @@ Reactions:
     <categoryEntry id="d684-5b63-caa7-adc4" name="Legiones Hereticus" hidden="false"/>
     <categoryEntry name="Forge Remnants - Militia Krios Squadron" hidden="false" id="dfbc-8edd-e14f-fd80"/>
     <categoryEntry name="Primus Medicae" hidden="false" id="b81a-8f4a-ff50-ef99"/>
+    <categoryEntry name="Tech-Priest Auxilia" hidden="true" id="86dd-51d8-3669-68ca"/>
```
nstephenh, line 1309: I set this to hidden, it works in nr but didn't appear to do anything in battlescribe.

2022 - Mechanicum.cat
```diff
@@ -2030,7 +2044,7 @@
                 <constraint field="selections" scope="parent" value="0" percentValue="false" shared="true" includeChildSelections="false" includeChildForces="false" id="8fe0-02cf-f34d-c225" type="min"/>
               </constraints>
               <entryLinks>
-                <entryLink id="6940-130e-f973-8451" name="Servo-Arm" hidden="false" collective="false" import="true" targetId="4168-fc85-8912-7188" type="selectionEntry">
+                <entryLink id="6940-130e-f973-8451" name="Servo-Arm" hidden="false" collective="false" import="true" targetId="4168-fc85-8912-7188" type="selectionEntry" defaultAmount="1">
```
nstephenh, line 2047: I also set this, doesn't do anything in battlescribe because the list that sets if you can take a servo-arm doesn't auto-set but in nr it sets to the first option.

nstephenh (approved):
Approving but letting you give it the once-over first as I made some minor changes.

PR #3080, opened 2024-01-17
Cleanup duplicates

nstephenh:
I wrote a script to check for all the duplicate entries. Here's the results. I'll re-run the script and re-save all the files tomorrow, as I think my branch was a bit old.
The things I have left to check before undrafting are:
- [x] ensure I'm not making [links to links](https://github.com/nstephenh/BSCopy/issues/10)
- [x] ensure no constraints or modifiers were dropped
- [x] check all the dark blessings and figure out what the current status of those is, unmerge them by hand.
- [x] manually move a few special rules, namely open-topped.

### DIFF
--- a/Daemon Library.cat
+++ b/Daemon Library.cat
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<catalogue xmlns="http://www.battlescribe.net/schema/catalogueSchema" library="true" id="31bf-eb23-33a-3d65" name="Daemon Library" gameSystemId="28d4-bd2e-4858-ece6" gameSystemRevision="77" revision="2" battleScribeVersion="2.03" type="catalogue">
+<catalogue xmlns="http://www.battlescribe.net/schema/catalogueSchema" library="true" id="31bf-eb23-33a-3d65" name="Daemon Library" gameSystemId="28d4-bd2e-4858-ece6" gameSystemRevision="77" revision="3" battleScribeVersion="2.03" type="catalogue">
   <sharedSelectionEntryGroups>
     <selectionEntryGroup id="002f-0e40-9d49-2a22" name="The Ætheric Dominion (X)" hidden="false" collective="false" import="true">
       <constraints>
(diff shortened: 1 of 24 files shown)


The4D6:
The scope of this PR has gotten too large.

Could we please try to split this down either in to two PRS (one for spaces, one for dupes), or break it down file by file?

It's hard to review given that you appear to have made edits to every file in the game, and interpreting the references just in GitHub is very difficult. Even when looked at in the NR editor it is not immediately apparent which version has been kept and why.

nstephenh:
> Could we please try to split this down either in to two PRS (one for spaces, one for dupes), or break it down file by file?
Sorry, I forgot to check out a new branch. This should now be based on master, and the spaces one should be merged already.

> It's hard to review given that you appear to have made edits to every file in the game, and interpreting the references just in GitHub is very difficult. Even when looked at in the NR editor it is not immediately apparent which version has been kept and why.
I somewhat agree, I'll be checking this in pycharm on the raw XML.
The prioritization I use is "GST, cat library, first appearance" in that order.

nstephenh:
Confirmed I'm definitely deleting modifiers so I'll need to rebranch and fix that.

nstephenh:
I think this is clean enough for review at this point, let me know if you have any questions or spot anything of concern.

Daemon Library.cat
```diff
@@ -21,7 +21,7 @@ A unit which includes models with this special rule without a specific Ætheric
           </constraints>
           <rules>
             <rule id="4438-874e-901d-f86" name="Encroaching Ruin" publicationId="8775-88f5-cfdd-24f6" page="3" hidden="false">
-              <description>When a unit entirely composed of models with this special rule and the Daemon Unit Type fails a Morale check, it only suffers 1 automatic Wound with no Saves of any kind allowed instead of D3 Wounds as specified in the Daemon Unit Type. In addition, models with this special rule gain the Move Through Cover special rule.</description>
+              <description>When a unit entirely composed of models with this special rule and the Daemon Unit Type fails a Morale check, it only suffers 1 automatic Wound with no Saves of any kind allowed instead of D3 Wounds as specified in the Daemon Unit Type. In addition, models with this special rule gain the Move Through Cover special rule.</description>
```
The4D6, line 24: I thought we'd move nonbreaking spaces to their own PR.

If so, that should be there

The4D6 (changes requested):
Looks good, just needs the one spacing issue moved out in to it's own PR as previously discussed, once that's done I'm happy with it

Daemon Library.cat
nstephenh: NBSP was merged already. This change appears to be " " to "nbsp". These are ones I recopied from the book to ensure I had the most recent wording, and that's why they have nbsp in them. Fixing.

PR #3090, opened 2024-01-28
Fixed rapier gunner weapon limit

The4D6:
Was 15, now 18


The4D6:
Changed it from a static limit to a flexible limit that increases per rapier weapon

nstephenh (approved):
If there's a better way of doing this, I can't figure it out.

PR #3100, opened 2024-02-03
Fix for RoW Disappearing and Aetheric Dominion (Whole Army) appearing when not wanted

Mayegelt:
Fix for RoW Disappearing when you take an allied detachment (Basically moved the RoW selection from LA Lib to each individual legion list and turned off import) Aetheric Dominion (Whole Army) from the DOTRS appearing when not wanted in armies like Militia. Fixed by turning off Import.

### DIFF
--- a/2022 - LA - Sons of Horus.cat
+++ b/2022 - LA - Sons of Horus.cat
@@ -6368,7 +6368,7 @@ A Warlord with this Trait gains a bonus of +1 to its Leadership Characteristic w
         <categoryLink id="7254-ed68-3002-6891" name="New CategoryLink" hidden="false" targetId="d494-e450-d4aa-579a" primary="true"/>
       </categoryLinks>
       <entryLinks>
-        <entryLink id="ae0-abdc-31f7-210b" name="Rites of War (Text Only / Work in Progress)" hidden="false" collective="false" import="true" targetId="cf3e-1d75-6f91-651f" type="selectionEntryGroup">
+        <entryLink id="ae0-abdc-31f7-210b" name="Rites of War" hidden="false" collective="false" import="true" targetId="cf3e-1d75-6f91-651f" type="selectionEntryGroup">
           <modifiers>
             <modifier type="set" field="797b-0bd6-8cbf-f25a" value="0">
               <conditionGroups>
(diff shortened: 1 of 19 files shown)


Mayegelt:
Not sure why this has looked so messy on the commit...
I'll try again.

Mayegelt:
Ok mess fixed. Not sure why it put in a load of junk code. Probably due to me saving in BS then resaving in NR.

2022 - LA - Sons of Horus.cat
```diff
@@ -6358,4 +6358,34 @@ A Warlord with this Trait gains a bonus of +1 to its Leadership Characteristic w
...
+    <selectionEntry id="3b55-a060-d00a-20e7" name="Rite of War" hidden="false" collective="false" import="false" type="upgrade">
+      <constraints>
+        <constraint field="selections" scope="parent" value="1" percentValue="false" shared="true" includeChildSelections="false" includeChildForces="false" id="eb90-9235-aa1-ac41" type="min"/>
+        <constraint field="selections" scope="parent" value="1" percentValue="false" shared="true" includeChildSelections="false" includeChildForces="false" id="c6fd-e5cb-7f2d-154a" type="max"/>
+      </constraints>
+      <categoryLinks>
+        <categoryLink id="7254-ed68-3002-6891" name="New CategoryLink" hidden="false" targetId="d494-e450-d4aa-579a" primary="true"/>
+      </categoryLinks>
+      <entryLinks>
+        <entryLink id="ae0-abdc-31f7-210b" name="Rites of War (Text Only / Work in Progress)" hidden="false" collective="false" import="true" targetId="cf3e-1d75-6f91-651f" type="selectionEntryGroup">
```
The4D6, line 6371: Why is the category "Text Only / Work in Progress" ?

Mayegelt: > Why is the category "Text Only / Work in Progress" ?

That's an odd revert...
I think it's dating back from when the files were initially constructed. So NR has ported over the link with the text for the link as it used to be read when the entry link used to be links to the group selection entry which had that name...

Mayegelt:
Here is a picy of where the thing was copied from in the LA file :D

image

nstephenh (changes requested):
Instead of doing this, couldn't we just prevent militia from importing LA?

nstephenh:
Alternative proposed in #3101. I prefer mine, but will take this one if that's what y'all prefer.

Mayegelt:
> Instead of doing this, couldn't we just prevent militia from importing LA?

I am not sure why the LA file was importing militia i assume that was a reason for it.

Issue #3099, opened 2024-02-02
Solar Auxilla or Imperialis Militia minimum + Mech allied detach


giloushaker:
**Describe the bug**
The `Solar Auxilla or Imperialis Militia` category is only added on the `Imperialis Militia` entry if there is `Templar Assault (IF)`,
which (allegedly) is not required for a Mech allied detach

**To Reproduce**
1. Go to (https://www.newrecruit.eu/app/list/zWcuM)
2. Scroll and and click "Import List"
3. See error in the roster

**Expected behavior**
No Error

**Device and data:**
 App being used to create lists: New Recruit

**Additional context**
https://discord.com/channels/736936369670193332/1202756964325793822/1202765438446997564

The4D6:
This is an issue with NewRecruit, not with our data set, so it is being handled by them

giloushaker:
The warlord one is NewRecruit specific. i fail to see how this one is an issue with NewRecruit.

nstephenh:
@The4D6 I'm reopening this

The4D6:
@nstephenh @Mayegelt 

Please can you guys try to repro this issue in BattleScribe as I cannot

giloushaker:
Seems its just BattleScribe not showing that error, even though it should.

PR #3101, opened 2024-02-04
Move shared profiles to GST

nstephenh:
Instead of #3100 moving the rites of war, we can just not import the LA library into militia. This will fix rites of war from appearing in militia. (#3099 ?)
Keeps @Mayegelt 's fix for: Aetheric Dominion (Whole Army) from the DOTRS appearing when not wanted in armies like Militia. Fixed by turning off Import.


Mayegelt:
I didn't know why Militia was imported. So assumed there is reason for it.
If there is not then it seems ok to do as you suggested.

The4D6:
It was imported in order to use the same profiles for the land raider and rhino, so that of they were ever updated in the future we would only have to update one thing

Mayegelt:
Fair enough. So I guess either clone those 2 or move to GST or have a shared asset library.

nstephenh:
This PR moves them to the GST, apologies for not communicating that properly.

The4D6 (approved):
Seems like a reasonable fix, I didn't realise this would casue issues, sorry!

Issue #3113, opened 2024-02-13
Thousand Sons Khenetai Occult Cabal Blademaster has incorrect points of 5 for Meltabombs - should be 10


anthonylerpiniere:
**Describe the bug**
Thousand Sons Khenetai Occult Cabal Blademaster has incorrect points of 5 for Meltabombs - should be 10

**To Reproduce**
1. Add Khenetai Occult Cabal to a roster
2. Select Khenetai Blademaster
3. Meltabombs show as 5 points instead of 10

**Expected behavior**
Meltabombs for the Khenetai Blademaster should be 10 points

**Screenshots**
![Screenshot 2024-02-13 at 11 03 09](https://github.com/BSData/horus-heresy/assets/68685410/13e3c04e-51b9-49ef-b5bb-7cadeab4d631)

**Device and data (please complete the following information):**
 - OS: macOS Sonoma 14.2.1
 - App being used to create lists: Battlescribe
 - App version: 2.03.21
 - Data release version: 18.2.0, Thousand Sons v31

PR that fixes this issue: https://github.com/BSData/horus-heresy/pull/3114

PR #3114, opened 2024-02-13
Update 2022 - LA - Thousand Sons.cat

anthonylerpiniere:
## Fixed Units
Amend points cost for Khenetai Occult Blademaster Meltabombs from 5 to 10

### Related Issues

* closes # [3113](https://github.com/BSData/horus-heresy/issues/3113)


anthonylerpiniere:
I have no idea - I tried it twice and had the same result.  I've only updated the value for the meltabombs using Data Editor.  Something to do with the version of BattleScribe I have installed (2.03.21)?

anthonylerpiniere:
Saved in NR Editor and pushed changes.  The numerical changes have been sorted.  It looks like some of the other  elements have moved around in the file, but the values look to be the same.  The game system revision has been increased to 100 which isn't something I changed in the editor so not sure if that's a problem or not.

nstephenh:
> The game system revision has been increased to 100 which isn't something I changed in the editor so not sure if that's a problem or not.

Game system revision auto-updates from the revision number of the game system. That's fine.

nstephenh (approved):
While the other stuff moving around is annoying, it's probably just because the thousand sons cat hasn't been edited in a while and newrecruit has likely made changes to the saving code.

Issue #3153, opened 2024-03-28
Imperial Fist Tartaros Command squad unable to select Vigil Storm Shield


kd7svh:
**Describe the bug**
Imperial Fist Tartaros Comand Squad is missing the option to replace it's ranged weapon with a Vigil Storm Shield.

1. New Recruit -> Horus Heresy (2022) -> Imperial Fists
2. Add Tartaros Praetor.  This model can correctly select a Vigil Storm Shield ("Legion Specific Upgrade")
3. Under "Retinue" option, select "Command Squad, Tartaros"
4. Note that there is no option for Vigil Storm Shields.  This should be in the "Terminator Ranged Weapon" section as it replaces their combi-bolter (unlike with the Praetor)

**Expected behavior**
In the "Terminator Ranged Weapon" section there should be the option to replace the combi-bolter with a Vigil Storm Shield.

**Device and data (please complete the following information):**
 - OS: MacOS
 - App being used to create lists: New Recruit
 - App version: v242
 - Data release version: Can't see an overall version but the Imperial Fists list states "rev 32"

**Additional context**
This all happened due to bug "Imperial Fists HQs Cannot Swap Guns for Shields #3069"

Where user The4D6 stated:
"I may be wrong here, so feel free to correct me if I am.
Both kinds of Imperial Fists terminator praetors have the option to exchange a combi-bolter for a vigil storm shield for 10pts.
They should not have this option."

He is wrong.  Very wrong.  Liber Astartes pg 228 clearly states that Imperial Fist Tartaros and Cataphractii units and command squads can replace their combi-bolter with a Vigil Storm Shield.  +15 points for Tartaros squads and +10 for Cataphractii.

Not to be a jerk, but it's really annoying that this got broken because some random files a bug stating what he THINKS is a problem and does zero actual investigation.  This is made worse by the fact you guys just took his word for it.  Hot tip, if a bug report starts with "I may be wrong" just assume they are.  Particularly if they aren't going to back their nonsense up with references.

Mayegelt:
Hi ya.
The issue has actually emerged due to something else being changed by the looks of it.
If you notice if you takeba command squad of termy with a regular praetor (not Cata or Tart) the option to replace the ranged weapon works fine.
However it seems when you select a Cata or Tart Praetor with the Cata or Tart Command squad the issue seems to happen.

It will be fixed again. Unfortunately this has happened a couple of times in exactly the same scenario of X + Y not allowing Z.
[...]

PR #3154, opened 2024-03-28
Vigil Storm Shields again and Sabre Pintle weapons

Mayegelt:
#3153
Issue happened again with the coding of the Termy weapon SGSE for vigil shields in Jan/Feb. This fix was put in to fix that Termy Praetors were mistakenly being given an option to swap their combi weapon for a vigil shield. However in fixing it it then locked out any command squad they took from taking them (though was ok for Artificer Praetors). It was a problem with "Ancestor Of" being used for the Praetors without making an exception for the command squad that comes below them.

I have made what seems to be a fix by splitting the "OR" for hidden from just Cata or Tart praetor as Ancestor to then have 2 "AND" sections of the same function, but then added 2 additional lines to also check to make sure they are "ancestor is not instance of cata / termy command squad" as well.

I've only tested on BS, and I know NR handles things differently.


Mayegelt:
Fix for Sabre Tanks Pintle Weapons.
There was an issue on BS that when you selected any of the combi weapon options for sabres, it was enforcing the selection on the whole group. This was down to the entries linked were for some reason linked to the "Collective" version of the weapons, and not the individual ones.

Mayegelt:
Strange things going on. It is saying over 50,000 changes, gunna try to work out why. This happened before and on resubmitting the changes it worked.
Edited in NR editor, so shouldn't have issues...

Mayegelt:
@nstephenh @The4D6 Ok. I have changed the bits I needed to. However as you might be able to see from the other pull request. It seems when I save (even if the only change I made was change the version number) it then throws a fix and changes 50,000 entry lines.
I am assuming this might be something with that code that was put in to remove spaces or removing the .0s or something...

PR #3155, opened 2024-03-28
Test of why it is making so many changes to file.

Mayegelt:
This is just a test...


Mayegelt:
@nstephenh @The4D6 
So no idea what is going on with this.
It seems when I save (even if the only change I made was change the version number) it then throws a fit and changes 50,000 entry lines.
I am assuming this might be something with that code that was put in to remove spaces or removing the .0s or something... This was done in NR editor. But something is clearly messed up somewhere, but no idea where or why.
A couple of these attempts are just downloading and reuploading the file. But as soon as that save button is hit after a change (even just the version number) it then says there are way more changes than there are.

Issue #3152, opened 2024-03-26
Veterans Counting Drop Pods 1-in-5


The4D6:
For the purposes of determining how many specialist weapons can be taken, Legion Veteran squads are counting their dedicated transport as a model in the unit.

This does not strike me as intended behaviour

nstephenh:
I can see a couple ways of solving this, both for this instance and long term.
### Option 1: 
We move units into a subunit inside of the unit. This also fixes count issues, where units display like this:
![image](https://github.com/BSData/horus-heresy/assets/3454308/d03a838b-a0bf-4a80-bc30-004888cc6645)
You can see an example on this branch/commit: https://github.com/BSData/horus-heresy/commit/5b5456c8a66173ba3240e4138744c14b615c8bfc
The one issue I forsee with this is if we need to then do a "number of models in the parent" check for "dedicated transport if 10 models or less" then we can't count number of models.
### Option 2:
We add a category for "models, excluding the transport. Example below: This we don't need to apply everywhere, we can do it ad-hoc.
https://github.com/BSData/horus-heresy/commit/cc55e1727c773e04f30a70c531f79653c7b81907
### Option 3:
Count models of a particular unit type (generally infantry) in the unit. 
This relies on us setting this properly for all model entries, which I don't think we've done. I might be able to do a script for it. This is the same as option 2 it just uses something pre-existing in the data.
### Option 4:
Both? Both is good. Do option 1 for the grouping improvements and option 3 for improved data + fix the issue with option 1.
### Option 5:
Change dedicated transports to "mounts" instead of models. This might break anything that counts number of models in force.

Mayegelt:
For me "Mounts" is a new thing I believe. Could a new new thing of "Dedicated Transport" be made for us?

Otherwise to me it seems easy enough to go for the option of "Count models in unit" as current. But with any with the option for dedicated transports, then include a "-1 if selection of (insert dedicated transport SEG)"

nstephenh:
> "Count models in unit" as current. But with any with the option for dedicated transports, then include a "-1 if selection of (insert dedicated transport SEG)"

We can't do that on the repeat, only on the final result. So it ends up looking like: (decrement max 0 should be decrement max 1)
![image](https://github.com/BSData/horus-heresy/assets/3454308/512db336-18c6-4c6a-8be3-87c770b37138)

[...]

PR #3160, opened 2024-03-28
Set types onto models based on their profiles

nstephenh:
This sets the categories on all models that have a profile or profile link as a direct descendent. This will allow us to check, for example, the number of infantry models in a given unit for #3152 .

It's also an improvement in data quality, and highlights some errors in the existing data like misspellings or missing unit types.


The4D6 (dismissed):
I can't seem to see any units that are missing form this change.

As long as we commit to applying this to future releases (eg Blackshields and Shattered Legions, though they use the marines units) then we should be ok

Issue #3135, opened 2024-02-27
Solar Auxilia New Units


nstephenh:
- [x] heavy sentinel: https://www.warhammer-community.com/wp-content/uploads/2024/02/ZwQB9kBXZA3CyHN4.pdf
- [x] light sentinel: to be released on warhammer community
- [x] Lifeward Section: to be released, in battle for beta-garmon

PR #3173, opened 2024-04-07
Beta garmon (because feature isolation is for chumps)

The4D6:
# Tasks:

### New Units

- [x] Endryd Haar

### Oaths
- [x] An Eternity of War
- [x] In Disgrace All Are equal
- [x] Only In Death Does Duty End
- [x] Panoply of Old
- [x] Pride is our Armour
- [x] The Broken Helix
- [x] The Eternal Vendetta
- [x] The Flesh Is Week
- [x] The Legacy of Nikaea
- [x] The Spoils of Victory
- [x] The Taint of Xenos
- [x] The Weapons of Desperation

# This PR does not validate properly in BattleScribe due to the lack of ability to check upstream values

Also on this branch:
Solar Aux New Units: #3135
 Horus Aximand - Sons of Horus
 Tybalt Marr - Sons of Horus
 Legion Optae - LA, Everybody can take!
 Legion Company Command Squad - LA, Everybody can take!

Recon Company

### DIFF
--- a/2022 - Horus Heresy.gst
+++ b/2022 - Horus Heresy.gst
@@ -2384,7 +2384,7 @@ If a Skimmer is Immobilised or Wrecked, its base is removed, if possible. If thi
                 <conditionGroup type="and">
                   <conditions>
                     <condition type="atLeast" value="1" field="selections" scope="roster" childId="ae4a-f95c-968e-eb46" shared="true" includeChildSelections="true"/>
-                    <condition type="instanceOf" value="0" field="selections" scope="ancestor" childId="d926-652f-8436-30ce" shared="true" percentValue="false" includeChildSelections="true" includeChildForces="false"/>
+                    <condition type="instanceOf" value="0" field="selections" scope="force" childId="d926-652f-8436-30ce" shared="true" percentValue="false" includeChildSelections="true" includeChildForces="false"/>
                   </conditions>
                 </conditionGroup>
               </conditionGroups>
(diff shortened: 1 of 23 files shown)


nstephenh:
This PR also has all the changes from my  units wave 1 PR so that should also be tested.
First suggestion, we should move blackshield oaths up in the xml so it's next allegiance/rite of war. 
We also talked about hiding rite of war and I don't think we did it, the change would be to uncheck the export flag (I think it's called import) in the blackshields cat and then manually import it in all the space marine cats.

2022 - Horus Heresy.gst
nstephenh, line 1: Two copies of blackshield primary and optional detachment

The4D6:
> This PR also has all the changes from my units wave 1 PR so that should also be tested. First suggestion, we should move blackshield oaths up in the xml so it's next allegiance/rite of war. We also talked about hiding rite of war and I don't think we did it, the change would be to uncheck the export flag (I think it's called import) in the blackshields cat and then manually import it in all the space marine cats.

I don't want to go making too many more changes, and I think messing about with Rites of War is outside the scope of this PR, they're all hidden from Blackshields anyway. Happy to make the change in a future piece of work however

Mayegelt:
The WoD update seems to be a roster miamer for Battlescribe users. It throws up errors on units using the "collective" option.
This isn't just on Blackshield armies, but impacts all legion armies. The only way to resolve seems to be deleting units or deleting collective models for those units and adding in a bunch of individuals like we have for Termys. This is a major knockback for QoL for Battlescribe users.
I can't really see a great alternative outside of remaking all the units for a Blackshields WoD variant of each of the units. It wouldn't be hard to copy paste each of them. But would of course mean that any updates that aren't legion specific (not including legion specific wargear that of course can be taken by both in theory) would need to also remember to be copied into the copy as well.

image

image

image

Test roster used for above picys as proof
https://cdn.discordapp.com/attachments/638905973481930773/1241446222384988293/IW_test.rosz?ex=664a3a64&is=6648e8e4&hm=450567f96eba6456fc269875ff7c795e4d34f4add16286e39e85e18fadb7bd9d&

nstephenh:
Legacy of Nikea, psycher sub-type is not selected by default, and the warp blast is 'hidden' but selected and that's throwing an error:
image

The4D6:
> Legacy of Nikea, psycher sub-type is not selected by default, and the warp blast is 'hidden' but selected and that's throwing an error: image

It shouldn't be selected by default, it's optional per-character

nstephenh:
Merging, we're just "Sending it"

PR #3192, opened 2024-05-19
Proof of concept Tactical WoD

Mayegelt:
Changes made.
Copy Legiones model entry.
Remove new part added of WoD that breaks old lists from original. Change name of copy of Legiones to Legiones (WoD) for reference only really, but might help. Set the defaults so it works better in NR as per original. Set original to be hidden if WoD is taken and copy to be hidden if not.

This does not fix the limitation on Heavy Stubbers, but there is a solution by individually adding it to the unit and linking it inside the GSE of each unit done unit by unit. If not a clarification note should probably be added that that particular restriction is not highlighted. Maybe adding to the name of Heavy Stubber something like "Heavy Stubber (WoD)*" and the * being referenced somewhere in the WoD rule to specify that the restriction needs to be checked manually, like we do in some of the RoW.


nstephenh:
I rebased this off the latest changes in the blackshields branch to mess with the differences.

nstephenh:
What's different about this compared to just using the legionary w options option?

PR #3209, opened 2024-05-28
Enforce max 3 on veletarii

nstephenh:
Using "Max 3 in self" doesn't work, so I've redone the logic per Giloushaker's suggestion. It will break existing lists in both editors I believe. Giloushaker can patch NR to fix it, but it'll still be a problem for battlescribe. We could also come up with a different validator, but this one is tricky. It's
Max 3:
0-1 veletaris command section
1-3 storm section, unless hermes, in which case 0-2
0-1 hermes, which modifies the storm section.
0-1 veletaris vanguard section

The other option would be to reduce max storm section for command section and vanguard section, like I'm currently doing with hermes. That should work...


nstephenh:
Doing the other option instead

PR #3217, opened 2024-05-29
Automated test suite

nstephenh:
To avoid regressions, I'm experimenting with automatically running tests


nstephenh:
Sounds like including roster files in the repo is not recommended, so I'll think of another solution such as renaming them.

nstephenh:
This now works, is independent from BScopy, and shouldn't break appspot. Any objections to merging, suggestions for other test cases, or general feedback?

nstephenh:
Merging this in, we can revert if if there's issues. I want to see how well the "only run tests on pull request or push to main" works, if you rebranch a pull request will it re-run for example?

PR #3228, opened 2024-05-31
Give Valdor the Primarch Force Org Slot

nstephenh:
Valdor had his primary set as "Primarch Unit Type" from when I redid the primarch unit types.


nstephenh:
Whoops, 4d6 did it already. Closing.

PR #3244, opened 2024-06-06
Order profiles

nstephenh:
Define profile ordering, for output.


Mayegelt:
Not sure what this is

The4D6:
Is this really necessary?

giloushaker:
Its for consistent profile ordering in the profiles popup & exports in NR
otherwise they will be in the order they are found in the data, and since they regroup profiles recursively they can be in an undesirable order

PR #3258, opened 2024-06-11
Overseers part 1

Mayegelt:
Overseers should work. Though might be a bit messy as I cloned the entry rather than linked to it, as I didn't want to catalogue link to both Militia and Solar cats


Mayegelt:
Had a little play with the new command squad thing to stop you taking them if you have another retinue in the form of praevian, mortificator or the overseers ones.

The4D6:
To clarify, the only centurion that should be able to take them is an un-upgraded one (ie one with no legiones consularis upgrade)

So you shouldn't be able to take them if you have any consularis, not just those that give "another" retinue

Mayegelt:
> To clarify, the only centurion that should be able to take them is an un-upgraded one (ie one with no legiones consularis upgrade)
> 
> So you shouldn't be able to take them if you have any consularis, not just those that give "another" retinue

So initial I thought this myself. However that was while I was under the impression that a Optae was a form of Consul.
However as it is not. It seems like it is just a generic command squad for centurions (as in the unit type, not the consularis selection).
I can easily change it back if it is not though.

The4D6:
The rule specifically says "Legion Centurion", and as per the boxout "Terminology: Centurions and Consuls" that refers only to a centurion with no Consul upgrade

the only units that are eligible to take a Company command Squad as a retinue are a Legion Centurion (no consuls at all) or the Optae

Mayegelt:
Ok cool. Will revert that change. Though I think it was still allowing it to be taken as other consul types on master I think.

nstephenh:
Cloning does prevent us from having to worry about any weirdness with provenance specific wargear.

nstephenh (approved):
I think this looks good, there's a couple of changes I don't quite know what there for but I'm good with this.

Issue #3247, opened 2024-06-06
Blackshields still show rite of war/redo imports in general


nstephenh:
I meant to fix this by changing how catalogue imports work but I apparently did not. 
Need to untick "import" on rite of war and/or change the import  on blackshields to not import root selection entries from LA.cat.

Mayegelt:
Moved the RoW to each legion in its own rite. This seems to have fixed the issue without causing probs for old rosters.

nstephenh:
I'll close this for now, we could mess around with imports more but I don't want to at this poin.

PR #3260, opened 2024-06-11
RoW change + Allied detachment change.

Mayegelt:
RoW moved from LA Lib into each legion individually without an Import flag.


Mayegelt:
#3247

Mayegelt:
Added in the Allied FoC change to make life better and more uniform. This will stop the issue of when making an allied detachment, that you have to then manually add the allied detachment SE to the force. It is now just underlinked to your Allegiance like ZM does to forces.

nstephenh (approved):
Looks good to me. Note that we probably need to make the release title something like "Reselect your rites of war or switch to new recruit" since battlescribe complains that it changes, but newrecruit handles it flawlessly.

PR #3266, opened 2024-06-14
Fix blackshields gaining white scars movement buff

### LINKED ISSUE #3218: Blackshields gaining White Scars movement buff
Blackshield units are gaining White Scars movement buff if you have taken Panopoly of Old Whitescars.
With the currently implementation every single unit will need multiple edits to fix I think as the buff is applied model by model, and in some cases in multiple ways such as a unit that can take jump packs or bikes or jetbikes

nstephenh:
This should fix #3218
Automated with bscopy, because I'd rather spend twice as long automating it then to do it by hand:
https://github.com/nstephenh/BSCopy/blob/main/find_conditions.py

There are 1122 modifiers using those conditions
20 modifiers already reference blackshields
1102 need review
89 of those appear to be whitescars movement

### DIFF
--- a/2022 - Mech Library.cat
+++ b/2022 - Mech Library.cat
@@ -6360,14 +6360,24 @@ A unit composed entirely of models with Utan jump boosters may not Run. During R
                 <modifierGroup type="and">
                   <modifiers>
                     <modifier type="increment" field="893e-2d76-8f04-44e5" value="1">
-                      <conditions>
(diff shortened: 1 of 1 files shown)


nstephenh:
That last commit got the last 4 movement modifiers I believe:
There are 1122 modifiers using those conditions
105 modifiers already reference blackshields
1017 need review
4 of those appear to be whitescars movement
4 conditions updated this run

nstephenh:
Merging this since it shouldn't cause any issues.

PR #3274, opened 2024-06-23
Typo's and Linebreaks 2

Mayegelt:
Fixes for a load more of the typos and linebreaks. This should get the majority in rules and wargear.


nstephenh (approved):
One minor thing, we should consider running a "fix link names" script in case anything referenced that rite of war.

Issue #3276, opened 2024-06-25
Sanguinary guard and sanctifier squads in alpha legion; turn off import all


nstephenh:
Sanguinary guard (blood angels) and sanctifier squads (salamanders) are showing up by default in alpha leegion
Per https://github.com/BSData/horus-heresy/issues/3247, might still be worth us messing around with "import root entries" as I don't know why that's set in alpha legion's imports of all the other cats.

PR #3277, opened 2024-06-26
Don't import all on other legions in AL.

nstephenh:
#3276

I don't know why alpha legion imported all root entries from other cats, but I've unchecked that setting. I tested at least one treachery unit inclusion, and it seemed to work. I assume we've added all the legion specific units as root entries to alpha legion, but if we did need this import, this will break some.


nstephenh:
Merging since @Mayegelt  agrees this should be fine.

PR #3278, opened 2024-06-26
Proof of concept for displaying reaction limit

nstephenh:
This currently looks really ugly because it displays the reaction limits as a cost, even though it's hidden. I've asked the NR devs to add a mega-ultra-hidden. The other way we could implement this is with a category instead of a cost... I think that would work much better, I'll consider redoing this.
image

### DIFF
--- a/2022 - Horus Heresy.gst
+++ b/2022 - Horus Heresy.gst
@@ -32,9 +32,9 @@
   </publications>
   <costTypes>
     <costType id="d2ee-04cb-5f8a-2642" name="Pts" defaultCostLimit="-1" hidden="false"/>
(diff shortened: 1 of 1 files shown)


Mayegelt:
How are you handling stuff like some of the special characters who give an extra in multiple phases if something happens or doesn't happen. Or restricting which reaction could be used?

nstephenh:
> How are you handling stuff like some of the special characters who give an extra in multiple phases if something happens or doesn't happen. Or restricting which reaction could be used?

I'm not currently, for those I imagine we could just add a specific override for "append *" by checking if the specific trait ID is in the roster.  Currently I just did the default warlord traits.

PR #3348, opened 2024-09-27
Shattered Test

Mayegelt:
Added a load of shattered legion stuff for test


Mayegelt:
Ok @nstephenh @The4D6 this is ready for a test. This should have everything in I think apart from 1 niggle that I can't really think how to put in the whole Major TS trait as it makes 1 guy (each phase) into a psyker. So do we edit every unit to have a Psyker*** thing with an explanation about it. OR just leave it up to them to play with.

PR #3524, opened 2025-01-25
correct the amount of attacks on command cadre bodyguard.

ReDTerraN:
In the miltia PDF it states that the bodyguard only has 1 attack in the base profile, shown here:

image

But the current data set shows two attacks as base, this is incorrect. I have adjusted it with this change.

## Added Units

Null

## Fixed Units

Command Cadre Bodyguard - Amount of base profile attacks

### Related Issues

Null

## Test Case

I dont know how to do that, but since its a minor change I think we should be good.


ReDTerraN:
Sorry for nerfing our already weak faction :upside_down_face:

The4D6 (approved):

PR #3560, opened 2025-02-14
Remove bolt pistol option for imperial militia wardens

ReDTerraN:
Miltia wardens cannot have bolt pistol option as shown here:
image

## Fixed Units
Imperialis Militia Rogue Psyker.

## Test Case

It has not been tested, but as its a minor change it should not be a big issue.


The4D6:
Please provide screenshots of this tested and working

PR #3561, opened 2025-02-14
Increase points of pintle heavy flamer on Imperial Militia Leman Russ

ReDTerraN:
As shown here:
image

for some god forsaken reason the heavy flamer costs 10 pts a piece.. Sorry for making this finding

## Fixed Units

Imperial militia Leman Russ

## Test Case

Not tested, but as its a minor change I dont think it will be a problem.


The4D6 (approved):
Looks good to me

PR #3708, opened 2025-06-28
Lots of fixes. Still lots to go before 3.0 hits.

Mayegelt:
Fix for logic on Bound Daemon Brutes for IT


Mayegelt:
Sang Guard in AL
Had to recode the Perdition weapons in sang guard, and added the unit entry to AL RoT

Mayegelt:
Custodian Guard weapon options lacked min/max 1 missing from the weapons and vex for the previews and stuff.

Mayegelt:
2 Special Char restrictions error
Removed the restriction from 2 chars as they do not have the loyalist faction, even though they probably should.

Mayegelt:
Shattered Legion missed Fury of Ancients Contemptor Talon and Venerable Dread root entries.
There might be more missing based on RoW.

